PR #1, opened 2018-06-19
Switch to released version of SmallRye Config

kenfinnigan:


antoinesd:
Version of this PR with passing build is here #2. Not sure that the config I changed is the reason...

PR #5, opened 2018-07-11
Introduce managed thread support

antoinesd:


pom.xml
```diff
@@ -39,7 +39,7 @@
     <version.junit>4.11</version.junit>
     <version.org.jboss.arquillian>1.1.13.Final</version.org.jboss.arquillian>
     <version.org.jboss.arquillian.container.weld-embedded>2.0.0.Beta5</version.org.jboss.arquillian.container.weld-embedded>
-    <version.org.jboss.logging>3.3.2.Final</version.org.jboss.logging>
+    <version.org.jboss.logging>2.0.1.Final</version.org.jboss.logging>
```
mkouba: Are you sure you want yo use this ancient version of jboss logging? What is the reason? Any compatibility issue?

antoinesd: No, it's only an alignment on version used in SmallRye Config: https://github.com/smallrye/smallrye-config/blob/master/pom.xml#L43

PR #15, opened 2018-09-11
 Make implementation more extensible and portable 

mkouba:


mkouba:
Looks like `maven-javadoc-plugin` settings are too strict...

antoinesd:
@mkouba since `maven-javadoc-plugin` config is inherited from SmallRye parent is it wise to override it in the project ?

mkouba:
Probably not, but it seems it wasn't so strict before, because I can see errors like:
```
[ERROR] /home/runner/smallrye-fault-tolerance/implementation/src/main/java/io/smallrye/faulttolerance/SimpleCommand.java:57: warning: no description for @param
[ERROR]      * @param operation
[ERROR]        ^
```

mkouba:
@antoinesd @kenfinnigan So it passed now ;-)

PR #21, opened 2018-09-19
Handle class hierarchies correctly

### LINKED ISSUE #20: GenericConfig does not handle class hierarchies correctly
If a method does not declare a FT annotation, the declaring type is always considered. In the following example the default value of `Retry#maxRetries()` is used even for `RetryOnClassServiceOverrideClassLevel`:
```java
@Retry
class BaseRetryOnClassService {
    void service() {
    }
}

@Retry(maxRetries = 4)
class RetryOnClassServiceOverrideClassLevel extends BaseRetryOnClassService {
}
```

mkouba:
- resolves #20


mkouba:
CI fails with usual TCK error. Pls review @antoinesd.

Issue #24, opened 2018-10-11
CommandListener should have a priority


mkouba:
So that it's possible to execute multiple listeners in predictable order.

We should also clarify which CDI scopes are supported and why.

PR #27, opened 2018-10-15
Add CommandListener.getPriority()

mkouba:
- also introduce CommandListenersProvider to work around limitation in
CDI API - we need to destroy dependent CommandListener instances after
execution is finished (this will allow integrators to leverage
non-portable APIs)


mkouba:
This should resolve #24.

mkouba:
@antoinesd CI build => intermittent TCK failure.

PR #46, opened 2018-11-21
Fix SynchronousCircuitBreaker - reset counters correctly

### LINKED ISSUE #28: Circuitbreaker not opening with certain parameters with fallback method
Hi,
I am trying to use the `CircuitBreaker` annotation with following parameters
```
@Fallback(fallbackMethod = "handleLoginFailure")
@CircuitBreaker(successThreshold = 1,requestVolumeThreshold = 2, failureRatio = 0.5,
delay = 15, failOn = {Exception.class}, delayUnit = ChronoUnit.SECONDS)
```
I make two sucessfull requests as expected and then it breaks as expected on the first unsuccessful request, but the subsequent request passes through as if it was closed. Am I missing something in the configuration or is this expected behaviour? Should'nt the delay be active now and the fallback method should be handling the subsequent requests during the duration of the delay?

/David

mkouba:
- i.e. whenever the requestVolumeThreshold is reached, even if the
status did not change
- resolves #28

### DIFF
--- a/implementation/src/main/java/io/smallrye/faulttolerance/SynchronousCircuitBreaker.java
+++ b/implementation/src/main/java/io/smallrye/faulttolerance/SynchronousCircuitBreaker.java
@@ -21,6 +21,7 @@ import static io.smallrye.faulttolerance.SynchronousCircuitBreaker.Status.OPEN;
 
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
+import java.util.LinkedList;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
@@ -50,13 +51,13 @@ class SynchronousCircuitBreaker implements HystrixCircuitBreaker {
         this.status = new AtomicReference<>(CLOSED);
         this.circuitOpenedAt = new AtomicLong(-1);
         this.successCount = new AtomicInteger(0);
-        this.failureCount = new AtomicInteger(0);
         this.halfOpenAttempts = new AtomicInteger(0);
         this.id = config.getMethodInfo();
         this.openTotal = new AtomicLong();
         this.halfOpenTotal = new AtomicLong();
         this.closedTotal = new AtomicLong();
         this.lastStatusChangeAt = new AtomicLong(System.nanoTime());
+        this.rollingWindow = new LinkedList<>();
     }
 
     @Override
@@ -109,12 +110,7 @@ class SynchronousCircuitBreaker implements HystrixCircuitBreaker {
                 return false;
             case OPEN:
                 if (isAfterDelay()) {
-                    LOGGER.debugf("OPEN >> HALF_OPEN [id:%s]", id);
-                    status.set(HALF_OPEN);
-                    halfOpenAttempts.set(1);
-                    long currentTime = System.nanoTime();
-                    openTotal.addAndGet(currentTime - lastStatusChangeAt.getAndSet(currentTime));
-                    resetCounts();
+                    toHalfOpen();
                     return true;
                 }
                 return false;
@@ -124,6 +120,7 @@ class SynchronousCircuitBreaker implements HystrixCircuitBreaker {
     }
 
     synchronized void executionSucceeded() {
+        record(true);
         successCount.incrementAndGet();
         Status current = status.get();
         if (HALF_OPEN == current && isSuccessThresholdReached()) {
@@ -136,7 +133,7 @@ class SynchronousCircuitBreaker implements HystrixCircuitBreaker {
     }
 
     synchronized void executionFailed() {
-        failureCount.incrementAndGet();
+        record(false);
         Status current = status.get();
         if (HALF_OPEN == current || (CLOSED == current && isFailureThresholdReached())) {
             // Transition to OPEN if HALF_OPEN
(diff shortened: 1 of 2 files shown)


mkouba:
In fact, I'm not really sure this PR is correct. I'm still confused. `CircuitBreaker#requestVolumeThreshold()` mentions a "rolling window". Does it mean we should always check the last `x` requests? And reset the window only if the circuit status changes?

mkouba:
In fact, I'm not really sure this PR is correct. I'm still confused. `CircuitBreaker#requestVolumeThreshold()` mentions a "rolling window". Does it mean we should always check the last `x` requests? And reset the window only if the circuit status changes?

mkouba:
@antoinesd pls review this PR carefully.

I've also created two related spec issues:
https://github.com/eclipse/microprofile-fault-tolerance/issues/342
https://github.com/eclipse/microprofile-fault-tolerance/issues/341

implementation/src/main/java/io/smallrye/faulttolerance/SynchronousCircuitBreaker.java
```diff
@@ -190,13 +199,27 @@ private boolean isHalfOpenAttemptAllowed() {
...
+    private void reset() {
+        successCount.set(0);
+        halfOpenAttempts.set(0);
+        rollingWindow.clear();
     }
 
-    private void resetCounts() {
-        successCount.set(0);
-        failureCount.set(0);
+    private int getFailureCount() {
```
michalszynkiewicz: It's probably practical to keep the window small so going through the lit shouldn't be a problem but if we wanted to be faster we could additionally keep counters as we did, just update them also based on the removed value.

mkouba: @michalszynkiewicz Yes, you're right. I think that a window will usually contain dozens of requests which shouldn't be a problem. Also I was thinking that in future we might need to store other attributes of a request, such as timestamp.

mkouba: By the way, how would you distinguish which result to remove if we only have counters? I.e. no ordering info.

michalszynkiewicz: I meant the list that we have now + counters

mkouba: I see!

implementation/src/main/java/io/smallrye/faulttolerance/SynchronousCircuitBreaker.java
```diff
@@ -207,10 +230,11 @@ private void resetCounts() {
 
     private final AtomicInteger successCount;
 
-    private final AtomicInteger failureCount;
-
     private final AtomicInteger halfOpenAttempts;
 
+    // failure = true, success = false
+    private final LinkedList<Boolean> rollingWindow;
```
michalszynkiewicz: Won't it misbehave with multiple threads? I think `ConcurrentLinkedQueue` should fit and be thread-safe.

testsuite/basic/src/test/java/io/smallrye/faulttolerance/circuitbreaker/rollingwindow/CircuitBreakerRollingWindowTest.java
```diff
@@ -0,0 +1,108 @@
...
+        } catch (IllegalStateException expected) {
+        }
+        try {
+            pingService.ping(true);
+            fail("Circuit should be open now");
+        } catch (CircuitBreakerOpenException expected) {
+        }
+        assertEquals(6, pingService.getPingCounter().get());
+
+        Thread.sleep(300);
```
michalszynkiewicz: what's the role of `sleep` here?

implementation/src/main/java/io/smallrye/faulttolerance/SynchronousCircuitBreaker.java
mkouba: Well, it's only manipulated from within synchronized methods. But `ConcurrentLinkedDeque` might be a better fit.

testsuite/basic/src/test/java/io/smallrye/faulttolerance/circuitbreaker/rollingwindow/CircuitBreakerRollingWindowTest.java
mkouba: `@CircuitBreaker(delay = 300)`

implementation/src/main/java/io/smallrye/faulttolerance/SynchronousCircuitBreaker.java
Ladicek: Doesn't the rolling window have a fixed size? Array-backed ring buffer should perform better than a linked list.

mkouba: It has and it should. But I don't think it makes sense to save few ms here.

Ladicek: The difference probably isn't even in millis :-) I actually think that plain `ArrayList` outperforms a `LinkedList` in almost all cases, but don't really mind.

mkouba:
@antoinesd I think we're ready to merge ;-)

PR #51, opened 2018-11-30
fallback on `@Retry.maxDuration` exceeded, fixes #39

### LINKED ISSUE #39: @Retry with timeout and @Fallback throws TimeoutException when execution time exceeded maxDuration
See also https://issues.jboss.org/browse/THORN-2196

michalszynkiewicz:


implementation/src/main/java/io/smallrye/faulttolerance/SimpleCommand.java
```diff
@@ -85,15 +96,38 @@ protected Object run() throws Exception {
...
+        if (retryContext == null || !retryContext.shouldRetryOn(failure)) {
+            return fallback.get();
+        } else {
             return super.getFallback();
         }
-        return fallback.get();
+    }
+
+    private Throwable translateException() {
+        Exception e = executionResult.getExecutionException();
```
michalszynkiewicz: this isn't very good.

testsuite/basic/src/test/java/io/smallrye/faulttolerance/bulkhead/BulkheadTest.java
```diff
@@ -50,22 +53,24 @@ public static JavaArchive createTestArchive() {
...
         int loop = QUEUE_SIZE * 2;
-        CountDownLatch startLatch = new CountDownLatch(loop);
+        CountDownLatch startLatch = new CountDownLatch(QUEUE_SIZE);
         CountDownLatch endLatch = new CountDownLatch(1);
         List<Future<String>> futures = new ArrayList<>();
         for (int i = 0; i < loop; i++) {
             futures.add(pingService.ping(startLatch, endLatch));
         }
-        startLatch.await(500, TimeUnit.MILLISECONDS);
+        assertTrue(startLatch.await(1000L, TimeUnit.MILLISECONDS));
```
mkouba: I think we should probably use something like this instead:
```java
startLatch.await(500, TimeUnit.MILLISECONDS);
Thread.sleep(500);
``` 
The reason is that `startLatch.await()` can eventually return before the waiting queue is full.

implementation/src/main/java/io/smallrye/faulttolerance/SimpleCommand.java
mkouba: Then we should probably leave a comment and create an issue...

michalszynkiewicz: done

testsuite/basic/src/test/java/io/smallrye/faulttolerance/bulkhead/BulkheadTest.java
michalszynkiewicz: done

PR #62, opened 2019-02-15
FT 2.0

michalszynkiewicz:


implementation/src/main/java/io/smallrye/faulttolerance/SecurityActions.java
```diff
@@ -17,43 +17,45 @@
...
+import java.util.function.Predicate;
+
+import static java.util.Arrays.asList;
 
 /**
- *
  * @author Martin Kouba
  */
-final class SecurityActions {
+public final class SecurityActions {
```
michalszynkiewicz: remove `public`

Ladicek: Fixed in https://github.com/michalszynkiewicz/smallrye-fault-tolerance/pull/13

Ladicek:
I triggered the build again and it passed: https://semaphoreci.com/smallrye/smallrye-fault-tolerance/branches/pull-request-62/builds/15

PR #65, opened 2019-03-06
fixes to get the MP RestClient integration working

Ladicek:


implementation/src/main/java/io/smallrye/faulttolerance/HystrixCommandInterceptor.java
```diff
@@ -169,7 +169,7 @@ public HystrixCommandInterceptor(@ConfigProperty(name = "MP_Fault_Tolerance_NonF
     public Object interceptCommand(InvocationContext invocationContext) throws Exception {
 
         Method method = invocationContext.getMethod();
-        Class<?> beanClass = interceptedBean != null ? interceptedBean.getBeanClass() : invocationContext.getTarget().getClass();
+        Class<?> beanClass = interceptedBean != null ? interceptedBean.getBeanClass() : method.getDeclaringClass();
```
Ladicek: I'm specifically interested in your comments on this change. I _think_ `interceptedBean` is only `null` in those out-of-spec cases, such as our MP RestClient integration, and so this change shouldn't affect the common usecases.

When I was running the tests for FT + RestClient integration, `invocationContext.getTarget()` was the proxy for the RestClient interface, which seems wrong to me. This change is an obvious fix, but not sure about all the consequences.

michalszynkiewicz: at first sight it looks okay. Plus the TCK tests for this are quite extensive.

PR #67, opened 2019-03-22
fix circuit breaker metrics in case an exception isn't considered a failure

### LINKED ISSUE #66: `circuitbreaker.callsFailed.total` metric incremented if an exception is thrown, even if it isn't present in `failOn`
If I have a `@CircuitBreaker(failOn = {IllegalArgumentException.class})` method and e.g. `IllegalStateException` is thrown, the `circuitbreaker.callsFailed.total` metrics currently gets incremented. It shouldn't be; instead, what should be incremented is `circuitbreaker.callsSucceeded.total`.

The MP Fault Tolerance spec is a little vague in this regard:

1. `circuitbreaker.callsSucceeded.total` is defined as "Number of calls allowed to run by the circuit breaker that _returned successfully_"
1. `circuitbreaker.callsFailed.total` is defined as "Number of calls allowed to run by the circuit breaker that then failed"
1. the `@CircuitBreaker` javadoc says "A method call will be considered a failure if it throws an exception and the type of that exception is assignable to any of the types listed in `failOn`."

From item 1 follows that in case of any exception, it should be the `circuitbreaker.callsFailed.total` metric that is incremented, because throwing an exception isn't a "successful return". However, from items 2+3 follows that it should be `circuitbreaker.callsSucceeded.total`.

The TCK of course doesn't test this. I'm going to file an MP Fault Tolerance issue for clarification, but my opinion is that the 2nd interpretation is correct, because that's how everything else in the circuit breaker works. Fix in SmallRye FT is pretty straightforward. 

FTR, this was originally filed as https://issues.jboss.org/browse/THORN-2361

Ladicek:
Resolves #66

### DIFF
--- a/implementation/src/main/java/io/smallrye/faulttolerance/HystrixCommandInterceptor.java
+++ b/implementation/src/main/java/io/smallrye/faulttolerance/HystrixCommandInterceptor.java
@@ -367,7 +367,7 @@ public class HystrixCommandInterceptor {
 
     }
 
-    private static Exception getCause(HystrixRuntimeException e) {
+    public static Exception getCause(HystrixRuntimeException e) {
         return (e.getCause() instanceof Exception) ? (Exception) e.getCause() : e;
     }
 
--- a/implementation/src/main/java/io/smallrye/faulttolerance/SynchronousCircuitBreaker.java
+++ b/implementation/src/main/java/io/smallrye/faulttolerance/SynchronousCircuitBreaker.java
@@ -258,7 +258,7 @@ public class SynchronousCircuitBreaker implements HystrixCircuitBreaker {
         }
     }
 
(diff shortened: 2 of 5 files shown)


Ladicek:
Draft for now, until https://github.com/eclipse/microprofile-fault-tolerance/issues/419 is resolved.

Ladicek:
Here's a PR that should resolve https://github.com/eclipse/microprofile-fault-tolerance/issues/419: https://github.com/eclipse/microprofile-fault-tolerance/pull/421 It was discussed yesterday on the MP FT call, so should be fine. I'll mark this ready when the spec+TCK PR is merged.

Ladicek:
Ready for review now, because the PR for clarifying the spec has been merged.

Ladicek:
Thanks!

PR #74, opened 2019-04-17
Propagate tracing context

pavolloffay:
Added per @kenfinnigan request https://github.com/quarkusio/quarkus/pull/2038#issuecomment-483636456

Signed-off-by: Pavol Loffay <ploffay@redhat.com>

### DIFF
--- a/implementation/src/main/java/io/smallrye/faulttolerance/HystrixInitializer.java
+++ b/implementation/src/main/java/io/smallrye/faulttolerance/HystrixInitializer.java
@@ -15,6 +15,9 @@
  */
 package io.smallrye.faulttolerance;
 
+import io.opentracing.util.GlobalTracer;
+import io.smallrye.faulttolerance.tracing.TracingConcurrencyStrategy;
+import io.smallrye.faulttolerance.tracing.TracingInstaller;
 import java.util.concurrent.TimeUnit;
 
 import javax.annotation.PostConstruct;
@@ -55,6 +58,7 @@ public class HystrixInitializer {
         LOGGER.info("Hystrix concurrency strategy used: " + strategy.getClass().getSimpleName());
         HystrixPlugins.getInstance().registerConcurrencyStrategy(strategy);
         HystrixPlugins.getInstance().registerCommandExecutionHook(new FaultToleranceCommandExecutionHook());
+        TracingInstaller.install();
     }
 
     @PreDestroy
--- a/implementation/src/main/java/io/smallrye/faulttolerance/HystrixInitializer.java
+++ b/implementation/src/main/java/io/smallrye/faulttolerance/HystrixInitializer.java
@@ -15,8 +15,6 @@
  */
 package io.smallrye.faulttolerance;
 
-import io.opentracing.util.GlobalTracer;
-import io.smallrye.faulttolerance.tracing.TracingConcurrencyStrategy;
 import io.smallrye.faulttolerance.tracing.TracingInstaller;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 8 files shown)


implementation/src/main/java/io/smallrye/faulttolerance/tracing/TracingInstaller.java
```diff
@@ -0,0 +1,25 @@
...
+
+  /**
+   * Install installs tracing if OpenTracing libraries are on classpath.
+   */
+  public static void install() {
+    try {
+      Class.forName("io.opentracing.Tracer");
+      TracingConcurrencyStrategy.register(io.opentracing.util.GlobalTracer.get());
+    } catch (ClassNotFoundException | LinkageError e) {
+      LOGGER.debug("OpenTracing is not on classpath, skipping context propagation instrumentation");
```
Ladicek: This should probably be an INFO. (I'm thinking WARN, but not sure.)

pavolloffay: Most of the cases tracing might not be present. I will change it to info then

Ladicek: Yea, I didn't think of that. So, hmm, I'm thinking, in case of Thorntail: to make this usable, the `io.smallrye.faulttolerance` module (as in JBoss Modules) would have to have a dependency on the `io.opentracing.tracer` module. But if OpenTracing isn't configured by some fraction, then it wouldn't do anything useful anyway. So we'd have to make the module dependency optional. Then, the `io.opentracing.tracer` module would only be present if the user depends on one of the following fractions: `opentracing`, `opentracing-tracerresolver` (but that's internal), `jaeger`, or `microprofile-opentracing`. Then it would work as intended...

In the end, it seems DEBUG was the right choice. Sorry about the fuss :-)

pavolloffay: This is pretty much implementation detail which should work without user interference. I will revert back to debug.

I am not fully following proposed thorntail impl. Can you make it work for thorntail?

implementation/src/main/java/io/smallrye/faulttolerance/HystrixInitializer.java
```diff
@@ -55,6 +58,7 @@ void onStartup() {
         LOGGER.info("Hystrix concurrency strategy used: " + strategy.getClass().getSimpleName());
         HystrixPlugins.getInstance().registerConcurrencyStrategy(strategy);
         HystrixPlugins.getInstance().registerCommandExecutionHook(new FaultToleranceCommandExecutionHook());
+        TracingInstaller.install();
```
kenfinnigan: Why not pass `HystrixPlugins.getInstance()` into the `TracingInstaller` so it doesn't need to reset all the plugins and re add them again?

implementation/src/main/java/io/smallrye/faulttolerance/tracing/TracingInstaller.java
Ladicek: Oh I can make it work with Thorntail no problem. I just had to go through this in my mind to understand how it's supposed to be used. Thanks for understanding :-)

implementation/src/main/java/io/smallrye/faulttolerance/HystrixInitializer.java
pavolloffay: The code was designed to work without passing strategies and plugins around. If we "own" the strategy installation we can go around without resetting. 

Is this code executed always even when a user provides custom concurrency strategy?

pavolloffay:
@kenfinnigan can we get this merged and cut a new release for me?

PR #81, opened 2019-05-09
fix @Asynchronous @Fallback with failed CompletionStage

### LINKED ISSUE #80: @Asynchronous @Fallback doesn't work with failed CompletionStage
If I have a method like this:

```java
@Asynchronous
public CompletionStage<String> hello() {
    CompletableFuture<String> future = new CompletableFuture<>();
    future.completeExceptionally(new RuntimeException());
    return future;
}

public CompletionStage<String> fallback() {
    return completedFuture("Fallback");
}
```

I expect that `hello()` returns a `CompletionStage` that will eventually complete with `"Fallback"`. That, however, isn't the case now. It eventually fails with `RuntimeException`.

Ladicek:
Fixes #80


Ladicek:
@mkouba FYI

PR #90, opened 2019-05-16
tone down logging for Hystrix concurrency strategy

Ladicek:


Ladicek:
This is a follow-up to #82.

PR #95, opened 2019-05-20
update to MP Metrics 2.0

### LINKED ISSUE #93: update to MP Metrics 2.0
MP Fault Tolerance integrates with MP Metrics. MP Metrics 2.0 has breaking changes. We need to adapt our code to these changes. Such change IMHO deserves a more significant version bump than a micro; at least 2.1.0, or the move to new coordinate scheme.

Ladicek:
Resolves #93

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,12 +38,12 @@
     <version.javax.interceptor-api>1.2.2</version.javax.interceptor-api>
     <version.mp-fault-tolerance>2.0.1</version.mp-fault-tolerance>
     <version.microprofile-config-api>1.3</version.microprofile-config-api>
-    <version.microprofile-metrics-api>1.1.1</version.microprofile-metrics-api>
+    <version.microprofile-metrics-api>2.0.1</version.microprofile-metrics-api>
     <version.slf4j-simple>1.7.26</version.slf4j-simple>
     <!-- smallrye-config is purely used for testing -->
-    <version.smallrye-config>1.0.0</version.smallrye-config>
+    <version.smallrye-config-1.3>1.0.0</version.smallrye-config-1.3>
     <!-- smallrye-metrics is purely used for testing -->
-    <version.smallrye-metrics>1.1.3</version.smallrye-metrics>
+    <version.smallrye-metrics-2.0>1.1.0</version.smallrye-metrics-2.0>
     <version.opentracing>0.31.0</version.opentracing>
 
     <!-- Used by release plugin to define git tag -->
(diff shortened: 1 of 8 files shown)


Ladicek:
Draft because MP FT TCK needs to be updated: https://github.com/eclipse/microprofile-fault-tolerance/pull/430

jmartisk:
I was just attempting to do exactly this but then I saw that there's a PR already :) Good. It will be needed to be able to update Quarkus and Thorntail to Metrics 2.0.
Is this planned to be in smallrye-fault-tolerance-2.0 version 1.0.0?

Ladicek:
The issue was in the `2.0-1.0.0` milestone, but this PR wasn't. It is now. So yes, this will be in `2.0-1.0.0`.

Ladicek:
When that happens of course depends on MP FT itself. I'll bring it up on today's MP FT call.

pom.xml
```diff
@@ -36,14 +36,14 @@
...
+    <version.mp-fault-tolerance>2.1-SNAPSHOT</version.mp-fault-tolerance>
     <version.microprofile-config-api>1.3</version.microprofile-config-api>
-    <version.microprofile-metrics-api>1.1.1</version.microprofile-metrics-api>
+    <version.microprofile-metrics-api>2.0</version.microprofile-metrics-api>
     <version.slf4j-simple>1.7.26</version.slf4j-simple>
     <!-- smallrye-config is purely used for testing -->
     <version.smallrye-config>1.3.7</version.smallrye-config>
     <!-- smallrye-metrics is purely used for testing -->
-    <version.smallrye-metrics>1.1.3</version.smallrye-metrics>
+    <version.smallrye-metrics-2.0>1.0.0</version.smallrye-metrics-2.0>
```
kenfinnigan: This needs to be 1.1.0 now as it was released today

Ladicek: Good! I wanted to converge on SmallRye Metrics 2.0-1.1.0 anyway, since I hoped it would have eager metrics initialization and thread safety fixes (and indeed it does have both).

kenfinnigan:
Is there an FT release coming or do we update the impl to work with Metrics 2 and ignore TCK for now?

Was there a meeting this week? MP Minutes don't show a meeting since the end of April

Ladicek:
FT release should be coming in 2 weeks, see the Gitter room. In the last few months, FT meetings have been spotty and minutes weren't taken even when the meeting itself did occur.

PR #99, opened 2019-06-17
update to latest SmallRye Parent POM and apply SmallRye code style

Ladicek:


Ladicek:
Most of the time, the autoformatter actually helps (or at least doesn't make things worse), but sometimes, it really is incredibly stupid. I took care of the most glaring offenders, but there are probably places that are now worse than they used to be.

implementation/src/main/java/io/smallrye/faulttolerance/CompositeObservableCommand.java
```diff
@@ -15,40 +15,40 @@
...
 public class CompositeObservableCommand extends HystrixObservableCommand {
 
     public static HystrixObservableCommand<?> create(Callable<? extends CompletionStage<?>> callable,
-                                                     FaultToleranceOperation operation,
-                                                     RetryContext retryContext,
-                                                     ExecutionContextWithInvocationContext ctx,
-                                                     MetricRegistry registry,
-                                                     boolean timeoutEnabled,
-                                                     Supplier<Object> fallback) {
+            FaultToleranceOperation operation,
```
michalszynkiewicz: :( former was nicer. Maybe moving callable to a next line would make it a bit beter.

Ladicek: Yea, it's based on Eclipse :-(

I'm personally ambivalent on indenting formal parameters. I usually do this:

```java
public static HystrixObservableCommand<?> create(Callable<? extends CompletionStage<?>> callable,
        FaultToleranceOperation operation, RetryContext retryContext, ExecutionContextWithInvocationContext ctx,
        MetricRegistry registry, boolean timeoutEnabled, Supplier<Object> fallback) {
    ...
}
```

But I just couldn't go through it all and fix everything I didn't like.

implementation/src/main/java/io/smallrye/faulttolerance/CompositeObservableCommand.java
```diff
@@ -80,40 +80,36 @@ protected CompositeObservableCommand(Callable<? extends CompletionStage<?>> call
...
-        Observable<Object> observable = Observable.create(
-                subscriber -> {
-                    try {
-                        if (registry != null && retryContext != null && retryContext.hasBeenRetried()) {
-                            counterOf(metricsPrefix + MetricNames.RETRY_RETRIES_TOTAL).inc();
-                        }
-                        CompletionStage<?> stage = callable.call();
-                        if (stage == null) {
-                            subscriber.onError(new NullPointerException("A method that should return a CompletionStage returned null"));
+        Observable<Object> observable = Observable.create(subscriber -> {
```
michalszynkiewicz: does the auto-formatter move the argument up?

Ladicek: I did that myself, because otherwise the result #$%^ing horrible :-)

michalszynkiewicz: for me the old form read much easier

Ladicek: I personally much prefer

```java
Observable.create(subscriber -> {
    ...
});
```

over

```java
Observable.create(
    subscriber -> {
    }
);
```

but feel free to run the reformat locally to see the result with original code.

PR #121, opened 2019-10-03
handle MP_Fault_Tolerance_NonFallback_Enabled properly

### LINKED ISSUE #120: Bug with MP_Fault_Tolerance_NonFallback_Enabled
Due to https://github.com/eclipse/microprofile-fault-tolerance/issues/451 the fact the implementation doesn't handle `MP_Fault_Tolerance_NonFallback_Enabled` properly was being hidden.

With the TCK fix, setting `MP_Fault_Tolerance_NonFallback_Enabled` to `false` prevents configuration overrides from enabling specific annotations

kenfinnigan:
Fixes #120

### DIFF
--- a/implementation/src/main/java/io/smallrye/faulttolerance/HystrixCommandInterceptor.java
+++ b/implementation/src/main/java/io/smallrye/faulttolerance/HystrixCommandInterceptor.java
@@ -44,7 +44,6 @@ import javax.interceptor.Interceptor;
 import javax.interceptor.InvocationContext;
 
 import org.eclipse.microprofile.config.Config;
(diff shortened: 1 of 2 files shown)


Ladicek:
I'm just thinking if you perhaps compared strings instead of classes intentionally, because comparing classes failed in Quarkus? If that's the case, and also considering https://github.com/eclipse/microprofile-fault-tolerance/pull/450, I'd say there are some serious classloading bugs in Quarkus itself, as this simply must work.

kenfinnigan:
I've verified that the change still works on Quarkus, not sure why I didn't use the class directly.

In the future, it would be appreciated if you could ask about making changes to the PR before doing them yourself. Thanks

PR #123, opened 2019-10-11
Fallback validation should blow up if the fallback method isn't from …

### LINKED ISSUE #122: Fallback validation shouldn't allow to define fallback method on a subclass
As was pointed out to me [here](https://github.com/eclipse/microprofile-fault-tolerance/pull/455#discussion_r333945906), spec forbids to declare fallback methods on subclasses:

> 

    When fallbackMethod is used a FaultToleranceDefinitionException will be thrown if any of the following constraints are not met:

        The named fallback method must be on the same class, a superclass or an implemented interface of the class which declares the annotated method
        [...]
        The named fallback method must be accessible from the class which declares the annotated method

Currently, smallrye doesn't seem to check that and we just happen to pass the TCK because we actually blow up (with exactly the same exception) on validating the [superclass](https://github.com/eclipse/microprofile-fault-tolerance/blob/master/tck/src/main/java/org/eclipse/microprofile/fault/tolerance/tck/fallbackmethod/beans/FallbackMethodSubclassBeanB.java) alone.

manovotn:
…same class, super class or implemented interface.

Fixes #122 

The tests for this are the TCKs alone because with changes merged in https://github.com/eclipse/microprofile-fault-tolerance/pull/4556, smallrye now fails the test in question.

### DIFF
--- a/implementation/src/main/java/io/smallrye/faulttolerance/config/SecurityActions.java
+++ b/implementation/src/main/java/io/smallrye/faulttolerance/config/SecurityActions.java
@@ -32,8 +32,10 @@ import java.security.PrivilegedExceptionAction;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Predicate;
 
 // this must be kept identical to io.smallrye.faulttolerance.SecurityActions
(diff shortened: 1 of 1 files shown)


implementation/src/main/java/io/smallrye/faulttolerance/config/SecurityActions.java
```diff
@@ -98,10 +100,20 @@ private static Method doGetMethod(final Class<?> beanClass, final Class<?> decla
         ParametrizedParamsTranslator expectedParamsTranslator = prepareParamsTranslator(beanClass, declaringClass);
 
         ParametrizedParamsTranslator actualParamsTranslator = new ParametrizedParamsTranslator();
+
+        // gather all interface methods based on declaring class
+        Set<String> allInterfaceMethodNames = new HashSet<>();
+        for (Class<?> clazz : declaringClass.getInterfaces()) {
```
Ladicek: Should this perhaps include super-interfaces of `declaringClass.getInterfaces()`?

Ladicek:
Otherwise looking good.

implementation/src/main/java/io/smallrye/faulttolerance/config/SecurityActions.java
manovotn: I don't think spec allows that. Quoting:

> The named fallback method must be on the same class, a superclass or an implemented interface of the class which declares the annotated method

From `implemented interface of the class...` I assume it has to be on an interface that directly implemented by given class.

Ladicek: I guess that's probably yet another vague formulation in the spec, but fair enough.

manovotn: Could be the case, to be fair I am learning about FT spec "on the fly" so take what I say with a pinch of salt ;)

PR #129, opened 2019-10-17
Update to MP Metrics 2.1.0 and SRye Metrics 2.2.0

kenfinnigan:
- Exclude failing MP FT TCK tests that have binary incompatibility with MP Metrics API class


Ladicek:
So, I'm just thinking if we should perhaps bump a minor version in SmallRye FT with this...

kenfinnigan:
I'm fine with releasing a 2.1.0

Want me to bump the version?

PR #137, opened 2019-11-08
Fixes #136: Only initialize Hystrix once per runtime

### LINKED ISSUE #136: Only initialize Hystrix once per runtime
```
18:44:03,986 ERROR [org.jboss.threads.errors] (ServerService Thread Pool -- 78) Thread Thread[ServerService Thread Pool -- 78,5,ServerService ThreadGroup] threw an uncaught exception: 
org.jboss.weld.exceptions.WeldException: WELD-000049: Unable to invoke void io.smallrye.faulttolerance.HystrixInitializer.onStartup() on io.smallrye.faulttolerance.HystrixInitializer@46ddadcf
```

rhusar:


kenfinnigan:
I understand we don't want to inadvertently re-run the initializer in a single deployment.

But in what situation is a single `HystrixInitializer` being used for multiple deployments? Aren't they in different classloaders? i.e. each deployment has its own `@ApplicationScoped` `HystrixInitializer`?

rhusar:
That's the contract of @ApplicationScoped....

kenfinnigan:
What do you mean, that there's one per deployment?

If so, why are we incrementing the counter on each subsequent call to the initializer?

Ladicek:
In WildFly, Hystrix isn't in the deployment's classloader, but in a single module.

rhusar:
Yes, that contact means there is one per deployment.

The counter keeps track of the first and last deployment. Registering subsequent concurrency strategy yields IllegalStateException.

kenfinnigan:
I'm extremely wary of these types of changes that rely on specific classloader behavior that is different from other runtimes.

Has this change been verified to work on Thorntail and Quarkus?

Ladicek:
The classloading behavior is the same in Thorntail and WildFly, except in Thorntail, we always have exactly 1 deployment, so we don't have to care. I assume it's the same with Quarkus, but I didn't check.

Ladicek:
Actually, I'm thinking about it again and what if different deployments want to have different Hystrix configuration? Then this change would mean that the 1st deployment decides the Hystrix config for everyone, which is ... wrong.

Perhaps it would be best to keep it failing if more than 1 deployment want to use Hystrix at the same time?

kenfinnigan:
Hadn't thought of that, but yeah it would mean every deployment has the same Hystrix configuration irrespective of what they've configured, which wouldn't be good

rhusar:
Please review Hystrix implementation and remember uniqueness principle of class loading in Java. HystrixPlugins is singleton as this will always return the same class loader of that Hystrix module:

```
    /* private */ static HystrixPlugins create() {
        return create(HystrixPlugins.class.getClassLoader());
    }
```

This is not a problem in Thorntail since it does not support multiple deployments.

rhusar:
> Actually, I'm thinking about it again and what if different deployments want to have different Hystrix configurations? Then this change would mean that the 1st deployment decides the Hystrix config for everyone, which is ... wrong. Perhaps it would be best to keep it failing if more than 1 deployment want to use Hystrix at the same time?

@Ladicek But that's a problem with Hystrix – how can Hystrix have more configurations? I don't think it can, but I would love to be proven wrong.

kenfinnigan:
Maybe, but the issue is that are users going to know that when using WF, or are they going to be expecting to be able to deploy different applications to the same WF instance and have different Hystrix configuration?

Ladicek:
Yea, Hystrix is not designed for that. The only way to make it work would be to put Hystrix into the deployment's module.

In the meantime, I'm leaning towards detecting the situation and throwing a proper error message, instead of silently letting 2nd deployment use the 1st deployment's Hystrix config.

rhusar:
@kenfinnigan That's a matter of documentation. Users need to follow migration path (e.g. clearly we will not be prefixing properties with "swarm.") and will be explained that configuration is pertaining to the entire runtime.

kenfinnigan:
I wouldn't be convinced people are going to read that kind of documentation, so expect lots of complaints

rhusar:
> In the meantime, I'm leaning towards detecting the situation and throwing a proper error message, instead of silently letting 2nd deployment use the 1st deployment's Hystrix config.

You are describing non-existent problem. TT doesn't support multiple deployments and neither does Quarkus. Current WildFly implementation only applies globally scoped configuration from MicroProfile Config so that issue doens't exist.

Ladicek:
In that case, throwing an exception if the counter of deployments ever goes higher than 1 is also not an issue, right? ;-)

rhusar:
@Ladicek Incorrect – its a problem for WildFly.

Ladicek:
In that case, I'll need to look more into "Current WildFly implementation only applies globally scoped configuration from MicroProfile Config". If the entire approach to integrating MicroProfile into WildFly relies on "no more than 1 deployment", then sure. Can you point me to some place this is documented, JIRAs etc.?

rhusar:
> If the entire approach to integrating MicroProfile into WildFly relies on "no more than 1 deployment"

@Ladicek  That's an incorrect assumption. WF has no problem with multiple deployments. There are issues e.g. around OpenAPI what multiple deployments on the same host need to expose as a resulting document. It's the other way round, this is needed only for this specific MP FT SR implementation which uses Hystrix which is essentially singleton. I recommend reviewing MP Config spec how configuration pertains to class loading. I don't think any of the other spec implementations have this problem. If it helps anything, think about it same as e.g. deployments cannot bring in their own JGroups configuration as its configured by the application server.

> Can you point me to some place this is documented, JIRAs etc.?

More than happy to and more than happy to get feedback on the branch. For starters, see the "Adding MicroProfile Fault Tolerance 2.0 support to WildFly" email in the SR mailing list, issues are https://issues.jboss.org/browse/WFLY-12590 and https://issues.jboss.org/browse/EAP7-1347, for proposal see https://github.com/wildfly/wildfly-proposals/pull/250 and the branch is https://github.com/rhusar/wildfly/tree/WFLY-12590.

Ladicek:
OK, got it. I misunderstood some of your explanations. Basically the FT configuration is global for the entire application server, and there's no per-deployment FT configuration. I think that's a reasonable limitation for now, until we get a new FT implementation in place.

rhusar:
No problem. If by "FT configuration" you mean "Hystrix configuration" then yes. The `MP_Fault_Tolerance_Metrics_Enabled` is scoped to deployment, so that should work as expected per deployment and so should `MP_Fault_Tolerance_NonFallback_Enabled` (let me know if you believe otherwise).

Ladicek:
Ah, OK. And there's also the ability to override annotation attributes via MP Config, that's deployment-scoped as well?

There's one thing that needs attention: some Hystrix configuration actually limit what you can sensibly configure with MP FT. For example, if you limit Hystrix thread pool size to 20, then you can't really configure MP FT `@Bulkhead` to 40. If the MP FT config is deployment-scoped and the Hystrix config server-scoped, then the Hystrix config sets a limit for all applications. I think that kinda sorta makes sense (or at least isn't complete nonsense), but needs to be properly documented.

rhusar:
Yeah, all of that is spec compliant and must be (e.g. `microprofile-config.properties` is associated with the classloader of the deployment).

> There's one thing that needs attention: some Hystrix configuration actually limit what you can sensibly configure with MP FT. For example, if you limit Hystrix thread pool size to 20, then you can't really configure MP FT @Bulkhead to 40. If the MP FT config is deployment-scoped and the Hystrix config server-scoped, then the Hystrix config sets a limit for all applications. I think that kinda sorta makes sense (or at least isn't complete nonsense), but needs to be properly documented.

Agreed, but this is basically no different then it is in Thorntail. You need to configure them both accordingly. The current section on that was ported to WF docs.

Ladicek:
It's not resource sharing, it's config sharing, and yes, that's the difference I had in mind. I think it should be stressed out in the docs.

PR #151, opened 2019-12-09
Fixes #150: @retry(maxRetries = -1) never retries

### LINKED ISSUE #150: @Retry(maxRetries = -1) never retries
Violates spec behavior defined in https://github.com/eclipse/microprofile-fault-tolerance/blob/master/api/src/main/java/org/eclipse/microprofile/faulttolerance/Retry.java#L55

If method in MP FT deployment is annotated by @Retry(maxRetries = -1) then it should it should retry forever until the method returns without exception. However it does not work.

Discovered by @mnovak1

rhusar:


rhusar:
@Ladicek I am planning to add this test to the TCK – WDYT? Is that the common practice here?

Ladicek:
@rhusar That definitely makes sense, please do! Thanks!

PR #162, opened 2020-01-09
update SmallRye Parent and move to Jakarta dependencies

### LINKED ISSUE #157: move from Java EE artifacts to Jakarta EE artifacts
https://github.com/orgs/smallrye/projects/2#card-30494938

This most likely needs to happen both in 2.x and 4.x.

Ladicek:
Resolves #157.

### DIFF
--- a/implementation/pom.xml
+++ b/implementation/pom.xml
@@ -101,20 +101,27 @@
       <artifactId>hystrix-core</artifactId>
     </dependency>
     <dependency>
-      <groupId>javax.enterprise</groupId>
-      <artifactId>cdi-api</artifactId>
+      <groupId>jakarta.enterprise</groupId>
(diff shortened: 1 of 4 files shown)


Ladicek:
Hi @rhusar, would this be OK/not OK/don't care for you, from WildFly perspective?

Ladicek:
Gonna merge anyway :-)

rhusar:
@Ladicek Back from PTO today. Looking.

rhusar:
@Ladicek So my understanding is that this OK for us from WildFly perspective (though not 100% sure about backports to older versions of e.g. EAP but as long its still using old package names in the code, that should be fine).

Ladicek:
Cool, thanks!

PR #168, opened 2020-01-15
fix the dependencies to use Jakarta EE artifacts

Ladicek:


implementation/fault-tolerance/pom.xml
```diff
@@ -100,20 +99,27 @@
...
+        <!-- https://github.com/eclipse-ee4j/interceptor-api/issues/31 -->
+        <exclusion>
+          <groupId>jakarta.ejb</groupId>
+          <artifactId>jakarta.ejb-api</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
-      <groupId>javax.interceptor</groupId>
-      <artifactId>javax.interceptor-api</artifactId>
```
Ladicek: I removed Interceptors completely, as it comes transitively through CDI.

testsuite/basic/pom.xml
```diff
@@ -29,13 +29,6 @@
     <dependency>
       <groupId>io.smallrye</groupId>
       <artifactId>smallrye-fault-tolerance</artifactId>
-      <exclusions>
-        <!-- https://github.com/eclipse-ee4j/interceptor-api/issues/31 -->
-        <exclusion>
-          <groupId>jakarta.ejb</groupId>
-          <artifactId>jakarta.ejb-api</artifactId>
-        </exclusion>
-      </exclusions>
```
Ladicek: This is now excluded in the implementation artifact, so no longer need to be excluded in test suite.

testsuite/tck/pom.xml
```diff
@@ -38,13 +38,6 @@
     <dependency>
       <groupId>io.smallrye</groupId>
       <artifactId>smallrye-fault-tolerance</artifactId>
-      <exclusions>
-        <!-- https://github.com/eclipse-ee4j/interceptor-api/issues/31 -->
-        <exclusion>
-          <groupId>jakarta.ejb</groupId>
-          <artifactId>jakarta.ejb-api</artifactId>
-        </exclusion>
-      </exclusions>
```
Ladicek: This is now excluded in the implementation artifact, so no longer need to be excluded in test suite.

implementation/fault-tolerance/pom.xml
```diff
@@ -100,20 +99,27 @@
...
     <dependency>
-      <groupId>javax.enterprise</groupId>
-      <artifactId>cdi-api</artifactId>
+      <groupId>jakarta.enterprise</groupId>
+      <artifactId>jakarta.enterprise.cdi-api</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>jakarta.el</groupId>
+          <artifactId>jakarta.el-api</artifactId>
+        </exclusion>
```
Ladicek: This is also excluded in MP FT.

PR #178, opened 2020-01-23
implement MicroProfile Fault Tolerance 2.1

### LINKED ISSUE #146: implement MicroProfile Fault Tolerance 2.1 changes
That is:

- added `CircuitBreaker.skipOn`
- added `Fallback.applyOn` and `Falback.skipOn`
- `Retry.retryOn` and `Retry.abortOn` no longer ignore `Throwable.class` (should be enough to just remove `SetOfThrowables.withoutCustomThrowables` and replace its usages with `create`)
- behavior of overlapping `Retry.retryOn` and `Retry.abortOn` was specified (we should be fine, and if not, the TCK will catch it)

Ladicek:
Resolves #146.

### DIFF
--- a/implementation/fault-tolerance/src/main/java/io/smallrye/faulttolerance/FaultToleranceInterceptor.java
+++ b/implementation/fault-tolerance/src/main/java/io/smallrye/faulttolerance/FaultToleranceInterceptor.java
@@ -32,6 +32,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 
 import javax.annotation.Priority;
+import javax.enterprise.context.control.RequestContextController;
 import javax.enterprise.inject.Intercepted;
 import javax.enterprise.inject.spi.Bean;
 import javax.inject.Inject;
@@ -74,6 +75,7 @@ import io.smallrye.faulttolerance.core.timeout.ScheduledExecutorTimeoutWatcher;
 import io.smallrye.faulttolerance.core.timeout.Timeout;
 import io.smallrye.faulttolerance.core.util.SetOfThrowables;
 import io.smallrye.faulttolerance.internal.InterceptionPoint;
+import io.smallrye.faulttolerance.internal.RequestScopeActivator;
 import io.smallrye.faulttolerance.internal.StrategyCache;
 import io.smallrye.faulttolerance.metrics.MetricsCollector;
 import io.smallrye.faulttolerance.metrics.MetricsCollectorFactory;
@@ -108,6 +110,8 @@ public class FaultToleranceInterceptor {
 
     private final StrategyCache cache;
 
+    private final RequestContextController requestContextController;
+
     @Inject
     public FaultToleranceInterceptor(
             FallbackHandlerProvider fallbackHandlerProvider,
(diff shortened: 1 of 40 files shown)


implementation/fault-tolerance/src/main/java/io/smallrye/faulttolerance/FaultToleranceInterceptor.java
```diff
@@ -158,6 +164,10 @@ public Object interceptCommand(InvocationContext invocationContext) throws Excep
                 CompletableFuture<T> result = new CompletableFuture<>();
                 asyncExecutor.submit(() -> {
                     try {
+                        // the requestContextController.activate/deactivate pair here is the minimum
+                        // to pass TCK; for anything serious, Context Propagation is required!
+                        // TODO does this code work together with Context Propagation?
+                        requestContextController.activate();
```
Ladicek: I'd be mostly interested in your thoughts about this. I'm pretty sure the `Future`  case is fine, but here, I'm not sure.

pom.xml
```diff
@@ -34,7 +34,7 @@
     <version.hdrHistogram>2.1.9</version.hdrHistogram>
     <version.jakarta.enterprise.concurrent-api>1.1.2</version.jakarta.enterprise.concurrent-api>
     <version.jakarta.interceptor-api>1.2.5</version.jakarta.interceptor-api>
-    <version.mp-fault-tolerance>2.0.3</version.mp-fault-tolerance>
+    <version.mp-fault-tolerance>2.1-RC2</version.mp-fault-tolerance>
```
rhusar: 2.1 final is now out: https://github.com/eclipse/microprofile-fault-tolerance/releases/tag/2.1

Ladicek: Right. I'm just testing one solution for the request context activation in presence/absence of Context Propagation, if it works, this PR will be ready for merging today :-)

rhusar: Great, just don't forget to branch before that :-) if that's still the plan.

kenfinnigan: Why branch before release?

rhusar: From what I understood, SR FT 4.0.x is already used in Quarkus following FT 2.0 spec. Thus a SR FT 4.1.x would be following the FT 2.1 spec.

kenfinnigan: Ok, but why is a branch needed?

I don't believe anything other than Quarkus is using 4.0.x, so we don't need to retain it for anything

Ladicek: Yea, we don't need a branch for 4.0 at this point in time. We can always create one retroactively, if there's a need.

rhusar: In general what we do in MW is branches are created to support different versions used in different projects. So in this instance, in order to support that Quarkus version using the SR FT 4.0.0 a "4.0.x" branch is created. If you think there will never be a 4.0.1 then certainly branch is not necessary.

kenfinnigan: For SmallRye we've taken the opposite approach, to only create branches when needed.

The reason is that most often a new release comes out and is used in downstream projects before any updates to an older release would be needed.

In this particular case, as Quarkus will update to the new 4.1.x release, we don't need a 4.0.x branch

Ladicek:
This PR should be final now. It uses MP FT 2.1 final release and passes TCK both with and without Context Propagation. Yay :-)

pom.xml
rhusar: Ok, although it's confusing and might fly for projects like Quarkus, but not branches that will be used in EAP given it is supported for a long time.

kenfinnigan: Any SmallRye implementation used in EAP, won't be supported for the same length as regular EAP.

It's something we still need to figure out

PR #196, opened 2020-02-25
increase test coverage for some smaller classes

Ladicek:


implementation/core/src/main/java/io/smallrye/faulttolerance/core/async/FutureExecution.java
```diff
@@ -51,7 +50,8 @@ public V get() throws InterruptedException, ExecutionException {
 
             @Override
             public V get(long timeout, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
-                return task.get().get(timeout, unit);
+                // at worst, the timeout here could possibly be 2x the requested value
+                return task.get(timeout, unit).get(timeout, unit);
```
Ladicek: This is actually a bugfix for an issue I found when writing the test. If the `@Asynchronous` method itself blocks, then `get(long, TimeUnit)` wasn't able to return within given timeout, because we were doing an unlimited `get()` first.

Downside of the solution here is that the real timeout could, in the worst case possible, be 2x the requested value. Which probably isn't all that bad.

PR #201, opened 2020-03-03
bugfixes

### LINKED ISSUE #197: Unexpected behaviour @Fallback method calls with @CircuitBreaker
MP FT does not behave correctly in case when there is @CircuitBreaker, @Fallback, @Retry and @Asynchronous on service method. @CircuitBreaker does not get opened based on its configuration and @Fallback method is called with incorrect value. 

Test scenario:
* Deploy MP FT service with:
```java
    @Asynchronous
    @Retry(retryOn = IOException.class)
    @CircuitBreaker(failOn = IOException.class, requestVolumeThreshold = 5, successThreshold = 3, delay = 2, delayUnit = ChronoUnit.SECONDS, failureRatio = 0.75)
    @Fallback(fallbackMethod = "processFallback")
    public CompletionStage<MyConnection> retryCircuitBreaker(int counter) throws IOException {
        System.out.println("retryCircuitBreaker - called - " + counter);
        if (counter % 4 != 0) { // 3/4 requests trigger IOException
            System.out.println("retryCircuitBreaker - called - " + counter + " throwing exception");
            throw new IOException("Simulated IOException");
        }
        System.out.println("retryCircuitBreaker - called - " + counter + " success");
        return CompletableFuture.completedFuture(new MyConnection() {
            @Override
            public String getData() {
                return "Hello from @Retry @CircuitBreaker method" + counter;
            }
        });
    }

    private CompletionStage<MyConnection> processFallback(int counter) {
        System.out.println("processFallback - returned url counter: " + counter);
        return CompletableFuture.completedFuture(new MyConnection() {
            @Override
            public String getData() {
                return "Fallback Hello" + counter;
            }
        });
    }
```

* Sends 16 parallel requests. Every 4th requests pass (including 0th one as there is modulo 4), and 12 invocations fail on IOException.
* Try on more request and check that @Fallback method was called as circuit is open

Expected result:
Circuit gets opened and @Fallback method is called. 

Actual Result:
Circuit is not opened and @Fallback method is called. 

https://issues.redhat.com/browse/WFWIP-307

### LINKED ISSUE #198: Unexpected behaviour @Fallback method calls with @Bulkhead
MP FT does behave correctly in case when there is @Bulkhead, @Timeout and @Fallback and @Asynchronous on service method. Some of requests do not go to fallback method but end with InterruptedException.

Test scenario:

Deploy MP FT service with:
```java
    @Asynchronous
    @Bulkhead(value = 15, waitingTaskQueue = 15)
    @Timeout(value = 1000)
    @Fallback(fallbackMethod = "processFallback")
    public CompletionStage<MyConnection> bulkheadTimeout(boolean fail) throws InterruptedException {
        if (fail) {
            Thread.sleep(2000);
        }
        return CompletableFuture.completedFuture(new MyConnection() {
            @Override
            public String getData() {
                return "Hello from @Bulkhead @Timeout method";
            }
        });
    }

    private CompletionStage<MyConnection> processFallback(boolean fail) {
        return CompletableFuture.completedFuture(new MyConnection() {
            @Override
            public String getData() {
                return "Fallback Hello";
            }
        });
    }
```

Send 10+ parallel requests with fail == true
Expected result:
All requests should go to fallback.

Actual Result:
Random number of requests (close to the number of requests) go to fallback.

https://issues.redhat.com/browse/WFWIP-308

### LINKED ISSUE #199: fallback is always the same for given class.method, even if parameters are different
```java
@ApplicationScoped
public class HelloService {
    public String hello(int counter) throws IOException {
        throw new IOException();
    }

    public String fallback(int counter) {
        return "Hello " + counter;
    }
}
```

When a service like this is invoked with varying values of `counter`, the results must vary as well. However, I always get the same fallback value.

This is because the chain of fault tolerance strategies is cached together with the `InvocationContext` of the call, which includes the method parameters.

Ladicek:
Resolves #197.
Resolves #198.
Resolves #199.

### DIFF
--- a/implementation/core/src/main/java/io/smallrye/faulttolerance/core/circuit/breaker/CompletionStageCircuitBreaker.java
+++ b/implementation/core/src/main/java/io/smallrye/faulttolerance/core/circuit/breaker/CompletionStageCircuitBreaker.java
@@ -1,7 +1,8 @@
 package io.smallrye.faulttolerance.core.circuit.breaker;
 
+import static io.smallrye.faulttolerance.core.util.CompletionStages.failedStage;
+
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
 
 import org.eclipse.microprofile.faulttolerance.exceptions.CircuitBreakerOpenException;
@@ -48,11 +49,12 @@ public class CompletionStageCircuitBreaker<V> extends CircuitBreaker<CompletionS
 
     private CompletionStage<V> inClosed(InvocationContext<CompletionStage<V>> target, CircuitBreaker.State state) {
         try {
-            CompletionStage<V> result = delegate.apply(target);
+            CompletableFuture<V> result = new CompletableFuture<>();
 
-            return result.handle((val, error) -> {
+            delegate.apply(target).whenComplete((value, error) -> {
                 if (error != null) {
-                    throw onFailure(state, error);
+                    onFailure(state, error);
+                    result.completeExceptionally(error);
                 } else {
                     metricsRecorder.circuitBreakerSucceeded();
                     boolean failureThresholdReached = state.rollingWindow.recordSuccess();
@@ -60,22 +62,18 @@ public class CompletionStageCircuitBreaker<V> extends CircuitBreaker<CompletionS
                         toOpen(state);
                     }
                     listeners.forEach(CircuitBreakerListener::succeeded);
-                    return val;
+                    result.complete(value);
                 }
             });
+
+            return result;
         } catch (Throwable e) {
-            CompletionException failure = onFailure(state, e);
-            return failedCompletionStage(failure);
+            onFailure(state, e);
+            return failedStage(e);
         }
     }
 
-    private CompletionStage<V> failedCompletionStage(Throwable failure) {
-        CompletableFuture<V> result = new CompletableFuture<>();
-        result.completeExceptionally(failure);
-        return result;
-    }
-
-    private CompletionException onFailure(State state, Throwable e) {
+    private void onFailure(State state, Throwable e) {
         boolean isFailure = !isConsideredSuccess(e);
         if (isFailure) {
             listeners.forEach(CircuitBreakerListener::failed);
@@ -96,12 +94,6 @@ public class CompletionStageCircuitBreaker<V> extends CircuitBreaker<CompletionS
 
             toOpen(state);
         }
-
-        if (e instanceof CompletionException) {
-            return (CompletionException) e;
-        } else {
-            return new CompletionException(e);
-        }
     }
 
     private CompletionStage<V> inOpen(InvocationContext<CompletionStage<V>> target,
@@ -109,7 +101,7 @@ public class CompletionStageCircuitBreaker<V> extends CircuitBreaker<CompletionS
         if (state.runningStopwatch.elapsedTimeInMillis() < delayInMillis) {
             metricsRecorder.circuitBreakerRejected();
             listeners.forEach(CircuitBreakerListener::rejected);
-            return failedCompletionStage(new CircuitBreakerOpenException(description + " circuit breaker is open"));
+            return failedStage(new CircuitBreakerOpenException(description + " circuit breaker is open"));
         } else {
             long now = System.nanoTime();
 
@@ -124,24 +116,36 @@ public class CompletionStageCircuitBreaker<V> extends CircuitBreaker<CompletionS
 
     private CompletionStage<V> inHalfOpen(InvocationContext<CompletionStage<V>> target, CircuitBreaker.State state) {
         try {
-            CompletionStage<V> result = delegate.apply(target);
-            metricsRecorder.circuitBreakerSucceeded();
+            CompletableFuture<V> result = new CompletableFuture<>();
 
-            int successes = state.consecutiveSuccesses.incrementAndGet();
-            if (successes >= successThreshold) {
-                long now = System.nanoTime();
-                closedStart = now;
-                previousHalfOpenTime.addAndGet(now - halfOpenStart);
+            delegate.apply(target).whenComplete((value, error) -> {
+                if (error != null) {
+                    metricsRecorder.circuitBreakerFailed();
+                    listeners.forEach(CircuitBreakerListener::failed);
+                    toOpen(state);
+                    result.completeExceptionally(error);
+                } else {
+                    metricsRecorder.circuitBreakerSucceeded();
 
-                toClosed(state);
-            }
-            listeners.forEach(CircuitBreakerListener::succeeded);
+                    int successes = state.consecutiveSuccesses.incrementAndGet();
+                    if (successes >= successThreshold) {
+                        long now = System.nanoTime();
+                        closedStart = now;
+                        previousHalfOpenTime.addAndGet(now - halfOpenStart);
+
+                        toClosed(state);
+                    }
+                    listeners.forEach(CircuitBreakerListener::succeeded);
+                    result.complete(value);
+                }
+            });
+
             return result;
         } catch (Throwable e) {
             metricsRecorder.circuitBreakerFailed();
             listeners.forEach(CircuitBreakerListener::failed);
             toOpen(state);
-            return failedCompletionStage(e);
+            return failedStage(e);
         }
     }
 }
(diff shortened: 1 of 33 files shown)


Ladicek:
Also CC @rhusar @mnovak1

This is best reviewed separately, one commit after the other.

I tried to run the WildFly tests by EAP QE and there's a couple of failures:

```
[ERROR] Failures: 
[ERROR]   MultipleFaultToleranceModuleEarTest.testDeployTwoModuleEarLogsWarning:110 Log does not contain warning that Hystrix was already initialized by first module
[ERROR]   MultipleJarDeploymentTest.testDeploySecondJarLogsHystrixNoConfigChangeWarning:81 Log does not contain WARNING Log: WFLYMPFTEXT0002: Hystrix was already configured! Skipping configuration from deployment - but it should.
[ERROR]   UndeployDeployTest.testDeployTwoWarsLogsWarning:252
[ERROR]   UndeployDeployTest.testNonMPFTDeploymentDoesNotBlockRestartHystrix:225 1 expectation failed.
Response body doesn't match expectation.
Expected: a string containing "Hello from @Timeout method, context = foobar"
  Actual: Fallback Hello, context = foobar

[ERROR]   UndeployDeployTest.testUndeployDeployChangesHystrixConfiguration:145 1 expectation failed.
Response body doesn't match expectation.
Expected: a string containing "Hello from @Timeout method, context = foobar"
  Actual: Fallback Hello, context = foobar
```

Just looking at the names, it seems these tests verify behavior of the old implementation that is no longer present, but it would be great if you guys could verify. It's possible these are genuine bugs.

Also if you could run the MP FT TCK with WildFly with this changeset, that would be great as well. I need to look into updating the Quarkus extension with this, because I had to do some changes in the CDI extension here :-)

Ladicek:
In general, running as many tests as we can with this would be beneficial :-)

mnovak1:
> Just looking at the names, it seems these tests verify behavior of the old implementation that is no longer present, but it would be great if you guys could verify. It's possible these are genuine bugs.

Correct, the other test failures can be ignored as they were relevant just for Hystrix. @tterem has updated branch of the tests for non-hystrix SM FT. @tterem could you link it here, please?

tterem:
Hi @Ladicek,

here is my branch with updated tests:

 https://github.com/tterem/eap-microprofile-test-suite/tree/1428

Tests that were depending on hystrix are removed there. You can try to run your fix against it. Meanwhile I will run as well.

Ladicek:
Cool, thanks.

tterem:
I build your branch locally and I confirm that our MicroProfile testsuite is passing with these changes.

implementation/core/src/main/java/io/smallrye/faulttolerance/core/circuit/breaker/CompletionStageCircuitBreaker.java
```diff
@@ -124,24 +116,36 @@ private CompletionException onFailure(State state, Throwable e) {
...
+            CompletableFuture<V> result = new CompletableFuture<>();
 
-            int successes = state.consecutiveSuccesses.incrementAndGet();
-            if (successes >= successThreshold) {
-                long now = System.nanoTime();
-                closedStart = now;
-                previousHalfOpenTime.addAndGet(now - halfOpenStart);
+            delegate.apply(target).whenComplete((value, error) -> {
+                if (error != null) {
+                    metricsRecorder.circuitBreakerFailed();
```
michalszynkiewicz: Maybe worth pulling out the three lines here to a separate method?

Ladicek:
Very nice, thanks! I'm just doing the same, it took me a while :-)

rhusar:
@tterem Great, thanks for the update! I was testing the integration and everything is passing there too.

> [INFO] Tests run: 407, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 406.126 s - in TestSuite

Ladicek:
Awesomesauce!

(I just ran the QE test suite locally and it's passing for me too.)

rhusar:
Are you sure you need the version bump? Nobody cares about internal/private API change.

implementation/core/src/main/java/io/smallrye/faulttolerance/core/retry/CompletionStageRetry.java
```diff
@@ -31,10 +33,8 @@ public CompletionStageRetry(
         return doRetry(ctx, 0, runningStopwatch, null);
     }
 
-    public CompletionStage<V> doRetry(InvocationContext<CompletionStage<V>> target,
-            int attempt,
-            RunningStopwatch stopwatch,
-            Throwable latestFailure) {
+    public CompletionStage<V> doRetry(InvocationContext<CompletionStage<V>> target, int attempt,
```
michalszynkiewicz: was it the auto-formatter? :(

rhusar:
Moreover, @quarkusio will need this fixes anyway so the 4.0.x will be just not in use by anybody.

implementation/core/src/main/java/io/smallrye/faulttolerance/core/retry/CompletionStageRetry.java
rhusar: Looks like unintended and unnecessary change.

rhusar: But n.b. given the time press its moot to me.

Ladicek:
I'm not sure about the version bump, but I figured I'd bump a minor as it might affect integrations. Now thinking about it again, the internal API changes only required changing the `FaultToleranceInterceptor`, so they are fully contained and integrations don't need to know. I'll probably remove that version-bumping commit.

implementation/core/src/main/java/io/smallrye/faulttolerance/core/retry/CompletionStageRetry.java
Ladicek: This might have actually been me :-) The one line per param format is so inefficient with screen space :-)

implementation/core/src/main/java/io/smallrye/faulttolerance/core/circuit/breaker/CompletionStageCircuitBreaker.java
Ladicek: I'd prefer not. I tend to not create methods that can possibly only be called from a single place.

Ladicek:
Removed the version-bumping commit.

implementation/core/src/main/java/io/smallrye/faulttolerance/core/circuit/breaker/CompletionStageCircuitBreaker.java
michalszynkiewicz: you have exactly the same lines in 145-147

Ladicek: Ah, true. That's in fact the same for all `CompletionStage*` strategies, and something I'd like to address in another way (later): the `Invocation` strategy would be, in the async flow, wrapped into an exception-handling strategy, which would catch exceptions thrown by the `Invocation` and propagate them to the `CompletableFuture`. No other `CompletionStage*` strategy would have to do that, anymore.

Ladicek: Filed https://github.com/smallrye/smallrye-fault-tolerance/issues/202

michalszynkiewicz: :+1:

implementation/fault-tolerance/src/main/java/io/smallrye/faulttolerance/FaultToleranceInterceptor.java
```diff
@@ -136,100 +138,80 @@ public FaultToleranceInterceptor(
...
         }
     }
 
     @SuppressWarnings("unchecked")
-    private <T> T syncFlow(FaultToleranceOperation operation,
-            Class<?> beanClass,
-            InvocationContext invocationContext,
-            MetricsCollector collector,
-            InterceptionPoint point) throws Exception {
+    private <T> T syncFlow(FaultToleranceOperation operation, InvocationContext invocationContext,
```
michalszynkiewicz: this is much less readable than one arg per line

PR #211, opened 2020-03-31
fix the stress test for async fault tolerance strategies

Ladicek:


Ladicek:
This is a follow-up to #208.

Ladicek:
Some more info about why this is correct is in #212.

PR #216, opened 2020-04-09
Upgrade Eclipse MicroProfile Fault Tolerance API to 2.1.1

rhusar:


Ladicek:
This is something I need to test (both in SmallRye itself and with Quarkus) and, in case of issues, provide feedback in the MP FT community. But this won't be merged as is; we'll only update when final version of 2.1.1 is out.

rhusar:
@Ladicek Yeah that's fine - the intention is to see what CI thinks :-)

rhusar:
Great, looks like CI is happy with this (so does in WildFly).

Ladicek:
Actually I totally forgot to remove that workaround. I think let's keep this PR open and use it later for the real 2.1.1.

rhusar:
@Ladicek Yeah, I will update it immediately as a new release comes out. It will probably be just a retag of this RC.

Ladicek:
Most likely, yea.

rhusar:
RC2 is out!

Ladicek:
Thanks. I tried with RC2 locally earlier today, and also with Quarkus, TCK still passes :-)

Ladicek:
There's an RC3 now.

rhusar:
Right, I was already testing yesterday and it was passing locally. Lets see what CI thinks!

rhusar:
RC4 is out now.

rhusar:
Rebased to trigger new CI runs.

Ladicek:
Thanks. I've heard on the MP FT call yesterday that RC4 seems fine for everyone, so hopefully the final version will be out soon.

rhusar:
Final is out - updated the PR. The CI will probably fail initially since this will take a while to sync to maven repo.

Ladicek:
Yay! Thanks.

FYI, I plan to release 4.2.1 with this chane, and them bump to 4.3.0 and release that with #235.

rhusar:
Sounds good.

Issue #210, opened 2020-03-31
don't Thread.sleep in CompletionStageRetry


Ladicek:
If a retry has a configured delay, we always implement it using `Thread.sleep`. There's no other way in the synchronous scenario, but for an async (`CompletionStage`-based) one, we should be able to just defer the execution of "the rest" of the logic to the future, and not block one thread in the meantime.

This is a 2nd usecase for reimplementing thread scheduling logic, see also #209.

Ladicek:
Done in #300.

PR #218, opened 2020-04-20
use fault tolerance for CompletionStage-returning methods that are not @Asynchronous and therefore don't run on an extra thread

Ladicek:


Ladicek:
I didn't expect the TCK to test for this, but it does :-) Those failures would be expected at this point. We can make a switch that would be off by default.

Bigger concern is that we have this nasty bug: #210. I'm not sure we can merge this with #210 happening.

FroMage:
This seems good, but you actually have to consider the return types differently and support other types of reactive types. ATM the preferred way to do this is to use this sort of code:

```java
                ReactiveTypeConverter<Object> converter = null;
                if (ret instanceof CompletionStage == false
                        && (ret instanceof Publisher == false || ic.getMethod().getReturnType() != Publisher.class)) {
                    @SuppressWarnings({ "rawtypes", "unchecked" })
                    Optional<ReactiveTypeConverter<Object>> lookup = Registry.lookup((Class) ret.getClass());
                    if (lookup.isPresent()) {
                        converter = lookup.get();
                        if (converter.emitAtMostOneItem()) {
                            ret = converter.toCompletionStage(ret);
                        } else {
                            ret = converter.toRSPublisher(ret);
                        }
                    }
                }
                if (ret instanceof CompletionStage) {
                    ret = handleAsync(tm, tx, ic, ret, afterEndTransaction);
                    // convert back
                    if (converter != null)
                        ret = converter.fromCompletionStage((CompletionStage<?>) ret);

```

Which converts a non-CS async type into a CS. And if you're writing an interceptor, you have to convert it back to the original type afterwards.

This is using:

```xml
        <dependency>
            <groupId>io.smallrye.reactive</groupId>
            <artifactId>smallrye-reactive-converter-api</artifactId>
        </dependency>
```

FroMage:
This will get you support for `Uni` automatically :)

Ladicek:
We don't do other async types yet.

FroMage:
I know. I'm saying you should, for Quarkus ;)

Ladicek:
I just found we didn't have an issue for that yet. Filed #219.

Ladicek:
#307 does this properly, using the new `@Blocking` and `@NonBlocking` annotations.

PR #234, opened 2020-05-27
allow DefaultExecutorFactory subclasses to customize the pools thread factory

Ladicek:


rhusar:
Where is this coming from? The EE container will just implement `ExecutorFactory` as can be seen here https://github.com/wildfly/wildfly/blob/master/microprofile/fault-tolerance-smallrye/executor/src/main/java/org/wildfly/extension/microprofile/faulttolerance/FaultToleranceContainerExecutorFactory.java

Ladicek:
Sure -- I'm adding another way how to implement it in an EE container. By subclassing the default :-)

Ladicek:
Perhaps it's a bad idea...

rhusar:
Its not a problem, I was just wondering whether someone raised this problem or something..

Ladicek:
Nah, I'm updating to SmallRye Fault Tolerance 4.x in Thorntail and found I have to implement an `ExecutorFactory` myself. And the only thing I really want to modify is the thread factory for all the pools -- just like in WildFly. Everything else stays intact. So I figured I'd just subclass the default provided by SmallRye Fault Tolerance. Hence this PR.

Ladicek:
Thanks :-)

Ladicek:
I'm actually thinking perhaps that shouldn't be a method that can be overridden, but a parameter to a constructor. I'll take a few minutes to play with that, before I do the 4.2.1 release.

Ladicek:
Hmm, doesn't look like too big of a difference, so I'll keep it as is.

PR #242, opened 2020-05-29
Fixed release action to use next version.

radcortez:


Ladicek:
Not quite sure why I'm not in the SmallRye Fault Tolerance team, and @michalszynkiewicz isn't either, which is wrong :-)

radcortez:
You are now :)

Ladicek:
Indeed, thanks :-)

PR #264, opened 2020-08-04
implement MicroProfile Fault Tolerance 3.0

### LINKED ISSUE #185: rewrite metrics system
The metrics system we have mostly comes from SmallRye Fault Tolerance 2.x. We should be able to make it simpler; for example, SmallRye Metrics is now thread-safe, so we don't have to synchronize ourselves.

Ladicek:
There are 2 big changes in MicroProfile Fault Tolerance 3.0:

- Lifecycle of stateful fault tolerance strategies is now specified. The specification prescribes the exact behavior SmallRye Fault Tolerance had since the beginning. No implementation required.
- Metrics have been revamped. This is what this PR implements.

Resolves #185.


Ladicek:
We'll have to bump a minor or even a major version. At this point, I'm not sure which it should be. Any opinions @michalszynkiewicz @rhusar @radcortez @kenfinnigan?

Ladicek:
My personal opinion: from SmallRye Fault Tolerance perspective, the only breaking change we do is metrics. This doesn't sound like 5.0 material to me, so I'd probably lean towards making it 4.4, but I could easily be persuaded otherwise.

implementation/core/src/main/java/io/smallrye/faulttolerance/core/bulkhead/BulkheadBase.java
```diff
@@ -4,55 +4,16 @@
 
 import io.smallrye.faulttolerance.core.FaultToleranceStrategy;
 
-/**
- * @author Michal Szynkiewicz, michal.l.szynkiewicz@gmail.com
```
michalszynkiewicz, line 8: :(

implementation/core/src/main/java/io/smallrye/faulttolerance/core/fallback/FallbackEvents.java
```diff
@@ -0,0 +1,21 @@
+package io.smallrye.faulttolerance.core.fallback;
+
+import io.smallrye.faulttolerance.core.InvocationContextEvent;
+
+public class FallbackEvents {
+    public static class Defined implements InvocationContextEvent {
```
michalszynkiewicz, line 6: why not enum values? Sth like:
public enum FallbackEvents {
   DEFINED,
   APPLIED
}

implementation/core/src/main/java/io/smallrye/faulttolerance/core/metrics/GeneralMetricsEvents.java
```diff
@@ -0,0 +1,16 @@
+package io.smallrye.faulttolerance.core.metrics;
+
+import io.smallrye.faulttolerance.core.InvocationContextEvent;
+
+public class GeneralMetricsEvents {
+    public static class ExecutionFinished implements InvocationContextEvent {
```
michalszynkiewicz, line 6: again, seems to me that an enum would be a better fit

michalszynkiewicz:
:+1: from me to bump the minor
I hope to finish reviewing tomorrow

Ladicek:
Good point about enums, for some reason I didn't consider them, I don't know why. I'll try :-)

Ladicek:
Ah, I know why -- the event system is very primitive, it keys events by `Class` objects. If I used an enum, I'd have to have a `switch` inside the event handler, which would make the code harder to read.

implementation/core/src/main/java/io/smallrye/faulttolerance/core/metrics/GeneralMetricsEvents.java
Ladicek: Enum should actually work well here, I'll check.

Ladicek:
Actually thanks for suggesting enums, I can use the enum-singleton pattern to make the code much shorter.

Even though I can't really use

```java
public enum FallbackEvents {
    DEFINED,
    APPLIED;
}
```

I can at least use

```java
public class FallbackEvents {
    public enum Defined implements InvocationContextEvent {
        INSTANCE;
    }

    public enum Applied implements InvocationContextEvent {
        INSTANCE;
    }
}
```

which is much better than the original.

implementation/core/src/main/java/io/smallrye/faulttolerance/core/bulkhead/BulkheadBase.java
Ladicek: I tend to do this when I do larger modifications to the classes, such that I'd have to add another author tag for myself. I find them an unnecessary clutter that convey exactly zero information. Your authorship is carefully preserved in Git history :-)

Ladicek:
Force-pushed because I amended one of the commits with the enum-singleton stuff.

Ladicek:
I have manually reran the jobs and they finished successfully. I have no idea why they still show as "waitin for status" here in the PR.

radcortez:
Strange. For some reason it seems that the GH Actions didn't picked the new commit hash. Might be a bug. If everything is ok, just squash it in master that they should run.

Ladicek:
I've created a `next` branch so that I can merge my recent changes. They will stay in `next` at least until MP FT 3.0 is released. Until then, `master` is for 4.3.x development (there's been recently some bugfixes, so I might actually do a 4.3.1 release).

PR #277, opened 2020-09-01
force creating thread pools with SmallRye Context Propagation

Ladicek:
Latest version of SmallRye Context Propagation no longer creates
a new thread pool for `ManagedExecutor.builder().build()` when
a default thread pool is set (which it is in Quarkus, for example).

Currently, SmallRye Fault Tolerance requires new thread pools that
it can control, so this commit uses the SmallRye Context Propagation
specific API to force thread pool creation.

This unfortunately means that SmallRye Fault Tolerance is tied to
SmallRye Context Propagation. It was before, too, but that was
an accident and could be easily removed. It is an intention now.


Ladicek:
With the changes in `next` branch that get rid of all additional thread pools and require just 1, which an integrator should provide, this is no longer useful.

PR #279, opened 2020-09-03
rewrite thread pools handling

### LINKED ISSUE #278: more efficient implementation of thread pool style bulkheads
We have 2 implementations of thread pool style bulkheads: one uses its own thread pools, which is inefficient by definition, the other uses a shared thread pool and emulates a queue by submitting all tasks to the thread pool and letting those that should be queued wait on a semaphore, which is also terribly inefficient.

We need a single implementation (not sure if possible, there's a difference between `Future` and `CompletionStage` bulkheads, so at least a single _style_ of implementation) that will use a shared thread pool and will not put queued tasks into the thread pool.

### LINKED ISSUE #287: rework the thread pool integration
Right now, SmallRye Fault Tolerance insists on creating its own thread pools, which is a waste of resources. There's an integration point (`ExecutorFactory`) which integrators can provide, but it also encourages them to create new thread pools.

We need to rework this. Ideally, SmallRye Fault Tolerance would only use _one_ thread pool, and let the integrator provide it.

Related: #209 #278

Ladicek:
Resolves #278
Resolves #287


Ladicek:
This PR now contains a 2nd commit that overhauls the main thread pool integration point. It is now recommended for all integrators to provide their own thread pool (though there's a default). It's a breaking change for integrators.

Ladicek:
I didn't merge this yet because I want to write some stress tests.

Ladicek:
Tests added.

Ladicek:
Rebased, reimplemented `Future` bulkhead again, fixed some other issues, added configuration property for built-in executor queue size. Hopefully tests will pass now :-)

PR #291, opened 2020-09-29
introduce a new API for circuit breaker maintenance

### LINKED ISSUE #280: provide a way to reset circuit breaker state
At least for testing, it would be good to have a way to reset circuit breaker state. It might also be helpful for emergency maintenance tasks.

The API could look like this:

```java
interface CircuitBreakerMaintenance {
    CircuitBreakerState state(); // we already have CircuitBreakerState, let's just reuse it
    void reset();
}
```

Question is: when I want to `@Inject CircuitBreakerMaintenance`, how do I know which circuit breaker that is? We probably need to revive the idea of giving circuit breakers a name (from #69). Something like `@CircuitBreakerName("...")`, which can only be put on methods (even though `@CircuitBreaker` can be put on an entire class). 

Question is how this works with inheritance. Need to check!

(Something similar could be done for bulkheads, which is a 2nd stateful fault tolerance strategy we have.)

Ladicek:
The `CircuitBreakerMaintenance` interface provides access to current
circuit breaker state, as well as a method to reset the circuit
breaker to the initial (closed) state. It can be injected using
the `@CircuitBreakerName` qualifier.

The previous event mechanism for observing circuit breaker state
changes is removed, as well as the internal `CircuitBreakerListener`
API (which was only used for tests anyway).

Resolves #280.


Ladicek:
TODO: need to handle the situation where there are multiple methods with the same `@CircuitBreakerName`; also need to check if the situation can occur unknowingly due to inheritance

implementation/fault-tolerance/src/main/java/io/smallrye/faulttolerance/CircuitBreakerMaintenanceRegistry.java
```diff
@@ -0,0 +1,67 @@
...
+    void register(String circuitBreakerName, CircuitBreaker<?> circuitBreaker) {
+        registry.putIfAbsent(circuitBreakerName, circuitBreaker);
+    }
+
+    @Produces
+    @CircuitBreakerName("")
+    @Dependent
+    CircuitBreakerMaintenance produce(InjectionPoint injectionPoint) {
+        // TODO use injectionPoint.getQualifiers instead?
+        String name = injectionPoint.getAnnotated().getAnnotation(CircuitBreakerName.class).value();
```
Ladicek, line 29: Hi @manovotn, could you please advise what's best here? Would both `injectionPoint.getAnnotated().getAnnotation(...)` and `injectionPoint.getQualifiers()` work? Is there a Quarkus concern? Thanks!

manovotn: Hi @Ladicek 

From purely spec perspective, both are correct and Quarkus will allow you to do this as well. However, if you are paranoid then switch to `injectionPoint.getQualifiers()`.
The reason is that Quarkus doesn't have a full notion of `Annotated` and [creates it just ad-hoc](https://github.com/quarkusio/quarkus/blob/master/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/CurrentInjectionPointProvider.java#L77-L84) for this very use case (because we want to support `InjectionPoint` but we don't have the whole hierarchy of `Annotated` in ArC). Though like I said - this should work in Quarkus as well ;-)

Ladicek: Cool, thanks! I'll switch to `getQualifiers` then.

Issue #219, opened 2020-04-20
support additional async return types


Ladicek:
We should support additional asynchronous return types, in addition to just `CompletionStage`. That is:

- Mutiny: `Uni`
- RxJava: `Single`, `Maybe`, `Completable`
- Reactor: `Mono`

We can probably ignore stream types (`Multi`, `Observable`/`Flowable`, `Flux`) for now.

Ladicek:
See also https://github.com/smallrye/smallrye-fault-tolerance/pull/218#issuecomment-616513212

schulzp:
Hi @Ladicek!

I'd highly appreciate support for `Uni` and the like. It would save us the `@Asynchronous` overhead as we're already using Vert.X' asynchronous web client. Is this part of any roadmap, already?

Ladicek:
Hi @schulzp, I think what you're after is #218. This issue is just about allowing different return types for `@Asynchronous` methods. (Actually perhaps you want a combination of both :-) )

If you care about not having to use `@Asynchronous`, or more generally about its semantics of "always run this on an extra thread", you can comment on #218 and also on https://github.com/eclipse/microprofile-fault-tolerance/issues/533 -- hearing from community folks would be very good.

As for roadmap, unfortunately there's none at the moment, as I have very little time to spare on SmallRye Fault Tolerance :-(

Ladicek:
Done in #310.

PR #307, opened 2020-10-30
add support for @Blocking and @NonBlocking

Ladicek:


README.adoc
```diff
@@ -16,6 +16,45 @@ SmallRye Fault Tolerance is an implementation of {microprofile-fault-tolerance}[
...
+We also want to stay compatible with the MicroProfile Fault Tolerance specification as much as possible.
+For these reasons, SmallRye Fault Tolerance only considers these annotations for methods that use some fault tolerance strategy.
+
+===== Recommendation
+
+For methods that use fault tolerance and return `CompletionStage`, we recommend to declare their `@Blocking` or `@NonBlocking` nature.
+In such case, the `@Asynchronous` annotation becomes optional.
+
+We also recommend to avoid `@Asynchronous` methods that return `Future`, because the only way to obtain the future value is blocking.
+
```
Ladicek, line 57: Hi @cescoffier, could you please review this documentation piece? I'd like to know if this usage of `@Blocking`/`@NonBlocking` aligns with your usage and your expectations around these annotations. If the documentation misses something or is unclear, pointing that out would be great too :-) Thank you!

README.adoc
```diff
@@ -16,6 +16,45 @@ SmallRye Fault Tolerance is an implementation of {microprofile-fault-tolerance}[
...
+These annotations can have multiple meanings, depending on context.
+SmallRye Fault Tolerance only pays attention to these annotations if:
+
+* they are placed on methods that return `CompletionStage` (the `Future` type can't really be used for non-blocking processing);
+* they are placed on methods that apply some fault tolerance strategy (such as `@Fallback`, defined either on a method or a class).
+
+Under these circumstances, SmallRye Fault Tolerance assigns these annotations the following meaning:
+
+* `@Blocking` means that execution of the operation will be offloaded to another thread (that's what `@Asynchronous` does by default);
```
cescoffier, line 34: I would add something like: `Uses this annotation if you don't want to block the caller thread and the annotated method has blocking logic.`

cescoffier (approved):
It looks good to me. I've done a small suggestion, but I already more or less covered in the _recommendation_ section.

At some point, we would need to add `Uni` support ;-)

Ladicek:
OK, thank you, I'll incorporate that suggestion. There's #219 for tracking the `Uni` support. I have more time to work on SRye FT now, so it should happen in the coming months :-)

PR #313, opened 2020-11-11
add proper logging

### LINKED ISSUE #172: use JBoss Logging annotation processor for all of logging

### LINKED ISSUE #173: replace Tracer with proper trace logging
The codebase includes a `FaultToleranceStrategy` implementation called `Tracer` that can be put in between other `FaultToleranceStrategy`-ies to trace how FTStrategies are invoked. This should be replaced by proper trace logging.

Ladicek:
This includes 2 closely related changes:

1. Use JBoss Logging annotation processor to generate logger
   implementations from annotated interfaces. This gives us
   message codes and internationalization for free.
2. Add trace logging for fault tolerance strategies.

Resolves #172.
Resolves #173.


implementation/core/src/main/java/io/smallrye/faulttolerance/core/circuit/breaker/CircuitBreakerLogger.java
```diff
@@ -0,0 +1,56 @@
+package io.smallrye.faulttolerance.core.circuit.breaker;
+
+import org.jboss.logging.Logger;
+import org.jboss.logging.annotations.LogMessage;
+import org.jboss.logging.annotations.Message;
+import org.jboss.logging.annotations.MessageLogger;
+
+@MessageLogger(projectCode = "SRFTL", length = 5)
+public interface CircuitBreakerLogger {
+    CircuitBreakerLogger log = Logger.getMessageLogger(CircuitBreakerLogger.class,
```
rhusar, line 10: Given this is static, it should be `LOG`.

rhusar (changes requested):

implementation/core/src/main/java/io/smallrye/faulttolerance/core/async/AsyncLogger.java
```diff
@@ -0,0 +1,27 @@
...
+import org.jboss.logging.annotations.Message;
+import org.jboss.logging.annotations.MessageLogger;
+
+@MessageLogger(projectCode = "SRFTL", length = 5)
+public interface AsyncLogger {
+    AsyncLogger log = Logger.getMessageLogger(AsyncLogger.class, AsyncLogger.class.getPackage().getName());
+
+    @Message(id = 2000, value = "CompletionStageExecution started")
+    @LogMessage(level = Logger.Level.TRACE)
+    void completionStageExecutionStarted();
```
rhusar, line 14: This just doesn't work well. Typically TRACE/DEBUG are not internationalized as it was not a requirement in the i18n effort. There is little harm in doing it, however here it references class names so the usefullness of localizing this is questionable.

rhusar: Just to explain: the harm is the load this puts on the i18n teams.

implementation/core/src/main/java/io/smallrye/faulttolerance/core/circuit/breaker/CircuitBreakerLogger.java
Ladicek: It probably should, but it seems SmallRye project use lower-case in this case. See https://github.com/smallrye/smallrye/wiki/I18N-Messages-and-Logging or https://github.com/smallrye/smallrye-common/search?q=%22%40MessageBundle%22

I personally don't mind either way.

implementation/core/src/main/java/io/smallrye/faulttolerance/core/async/AsyncLogger.java
Ladicek: Ah, good point. I got a bit overzealous here :-) And I probably didn't even know. Thanks, I'll move trace logging to `log.trace("...")`.

Ladicek: Hmm, just found that e.g. SmallRye Config or Weld do the same thing -- DEBUG or TRACE logging via i18n interfaces. It's kinda natural to use it all the way through.

Given that the log level is defined statically, I think DEBUG and TRACE messages shouldn't even get to the l10n teams for translation, so the harm should really be none.

Just not sure what to do :-)

Ladicek: WildFly also seems to use i18n interfaces for DEBUG and TRACE logging, but that seems to be more of an accident, and vast majority of DEBUG and TRACE logging is directly in the code. So I'll do that too.

Ladicek: Done.

implementation/core/src/main/java/io/smallrye/faulttolerance/core/circuit/breaker/CircuitBreakerLogger.java
Ladicek: I also did this. All singleton logger instances are now called `LOG`, and are statically imported, leading to much nicer code.

PR #316, opened 2020-11-19
Publish TCK Report

radcortez:


rhusar (approved):
LGTM, unfortunately, it looks like there is no way to view the report from the UI...

radcortez:
> LGTM, unfortunately, it looks like there is no way to view the report from the UI...

No, it just attaches files. I guess that will be the next step and publish the files somewhere :)

Ladicek (approved):
Thanks!

Ladicek:
What I find a bit sad is that this isn't immutable, you can easily delete the file. But it's probably the best at the moment.

(I was thinking perhaps we could publish the result of the TCK run to Maven Central. That would abuse the purpose of Maven Central a bit, but would be quite a bit more reliable :-) )

Ladicek:
Actually I'm thinking, do we need it for all builds, or just the release build?

radcortez:
> What I find a bit sad is that this isn't immutable, you can easily delete the file. But it's probably the best at the moment.
> 
> (I was thinking perhaps we could publish the result of the TCK run to Maven Central. That would abuse the purpose of Maven Central a bit, but would be quite a bit more reliable :-) )

That may be an options. Ideally we should have a webpage where we could publish this results per build and make them immutable. Maybe we could use smallrye.io, but this was the easiest way to do it for now without too much effort :) 

> Actually I'm thinking, do we need it for all builds, or just the release build?

Probably just for the release build. I was playing with that to showcase it to the MP community and I didn't want to burn versions to have it only on release, so I add it on the regular build as well. Feel free to remove it :)

PR #321, opened 2020-11-25
Circuit breaker half open exceptions

### LINKED ISSUE #320: circuit breaker in half-open always treats exceptions as failures
When the circuit breaker is half-open, our implementation always considers exceptions as failures. This is wrong, the `failOn` and `skipOn` configuration needs to be considered, just like in the closed state.

Ladicek:
Fixes #320


Ladicek:
Wrongly created PR, need to open a new one.

PR #323, opened 2020-11-26
improve circuit breaker to reject excess attempts in half-open

### LINKED ISSUE #319: investigate handling of half-open circuit breakers
Currently, when we have a circuit breaker in the half-open state, we let all invocations go through and

- move back to open in case of any failure;
- move to closed after a configured number `successThreshold` of invocations passes.

This is fine and passes the TCK, but it defeats the purpose of circuit breakers under high load.

What we should be able to do is: only allow the first `successThreshold` invocations to go through, reject the rest. If any of those invocations fail, move back to open; if they all succeed, move to closed.

This matches the intent of circuit breaker much better, because under high load, only a few invocations will go through, most of them will still get rejected, avoiding a possible thundering herd on the guarded service.

Ladicek:
Previously, circuit breakers in the half-open state would let
all invocations go through. In case of any failure, it would
move back to open, and in case of a configured number of consecutive
successes, it would move to closed.

This is fine and passes the TCK, but under high load, this can
cause a thundering herd problem on the guarded service.

A better solution, implemented by this commit, is: in half-open,
only allow the first `successThreshold` invocations to go through
(also known as "probe invocations"), and reject the rest. If any
of those probe invocations fail, move back to open; if all of them
succeed, move to closed.

Fixes #319

### DIFF
--- a/implementation/core/pom.xml
+++ b/implementation/core/pom.xml
@@ -64,6 +64,45 @@
         </dependency>
     </dependencies>
 
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>test-jar</goal>
+                        </goals>
+                        <configuration>
+                            <includes>
+                                <include>io/smallrye/faulttolerance/core/util/barrier/*</include>
+                                <include>io/smallrye/faulttolerance/core/util/party/*</include>
+                            </includes>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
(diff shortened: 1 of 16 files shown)


Ladicek:
Weird, that test passes for me just fine locally. Need to investigate.

Ladicek:
OK, it does fail fairly often when I run the test repeatedly.

Ladicek:
Got it, it's a concurrency issue that causes an assertion to fail, which causes executor shutdown, which causes thread interruptions, which manifests as a weird test failure :-)

Ladicek:
Actually we've had the concurrency issue several times before, always solved it with using a pair of `CountDownLatch`es instead of just one. I got tired of that, so I built a simple abstraction over that called `Party` (yay! :balloon: :tada: :champagne:).

PR #353, opened 2021-01-12
Update to OpenTracing 0.33.

radcortez:


radcortez:
@Ladicek we will need this for the MP 4.0 update.

implementation/tracing-propagation/src/main/java/io/smallrye/faulttolerance/tracing/TracingContextProvider.java
```diff
@@ -41,9 +40,9 @@ public ThreadContextSnapshot clearedContext(Map<String, String> props) {
         return () -> {
             Tracer tracer = GlobalTracer.get();
             ScopeManager scopeManager = tracer.scopeManager();
-            Scope activeScope = scopeManager.active();
-            if (activeScope != null) {
-                activeScope.close();
+            Span span = scopeManager.activeSpan();
+            if (span != null) {
+                scopeManager.activate(span).close();
```
Ladicek, line 45: So this line is effectively a noop. It creates a new scope and immediately destroys it, which leads to restoring the previously active scope (if there was one). Looks like the new OpenTracing API doesn't let us do what ConProp requires :-/

Ladicek: ... which is _temporarily suspending the current **span**_. That would be done by _closing the currently active **scope**_ (and later activating the span again, which would create a new scope; we currently don't do this as highlighted by the TODO below). But we can't get to the _currently active **scope**_ anymore.

Ladicek: This actually seems to be a deliberate decision by OpenTracing to limit the API misuse. Which makes advanced use cases, like context propagation, impossible :-)

Ladicek: I suppose the OpenTracing API is meant to be used in a way that there is no _currently active **scope**_ (and hence no _currently active **span**_) by the time context propagation happens. Which essentially means that this method should be a real noop.

kenfinnigan: Not sure on this one, might be good to see if @pavolloffay can offer advice

radcortez: From the java doc:

```
    /**
     * @deprecated use {@link #activeSpan()} instead.
     * Return the currently active {@link Scope} which can be used to deactivate the currently active
     * {@link Span}.
     *
     * <p>
     * Observe that {@link Scope} is expected to be used only in the same thread where it was
     * created, and thus should not be passed across threads.
     *
     * <p>
     * Because both {@link #active()} and {@link #activeSpan()} reference the current
     * active state, they both will be either null or non-null.
     *
     * @return the {@link Scope active scope}, or null if none could be found.
     */
```

radcortez: My interpretation is that it doesn't necessarily create a new span. The implementation seems equivalent, but I'm no expert here :)

Ladicek: That javadoc is for `ScopeManager.active()` and comes from 0.32, right? Because that method is gone in 0.33.

`ScopeManager.activate(Span)` creates a new scope. The javadoc for `Span` and `Scope` could be clearer, but my understanding is that _scope_ is a slice of time during which you do something with the _span_. During the lifetime of a _span_, many _scopes_ can be created and subsequently closed, but the _span_ itself can only be closed once.

Ladicek:
Hmm, I wanted to force-push the commit I amended locally, but somehow failed. Sorry about that, will open a new PR from my branch :-)

Ladicek:
Ha, I know what happened, instead of pushing the commit, I accidentally pushed the `master` branch, and GitHub correctly figured out that there's nothing to merge, so it closed the PR, and now I can't push the commit to the branch, because it doesn't belong to a PR :-)

Ladicek:
Submitted #360.

implementation/tracing-propagation/src/main/java/io/smallrye/faulttolerance/tracing/TracingContextProvider.java
radcortez: > That javadoc is for `ScopeManager.active()` and comes from 0.32, right? Because that method is gone in 0.33.

Yes, it is.

PR #360, opened 2021-01-14
Update to OpenTracing 0.33

Ladicek:


radcortez (dismissed):

implementation/tracing-propagation/src/main/java/io/smallrye/faulttolerance/tracing/TracingContextProvider.java
```diff
@@ -12,43 +12,44 @@
...
 
-        if (activeScope != null) {
-            Span span = activeScope.span();
+        if (span != null) {
             return () -> {
-                Scope propagated = scopeManager.activate(span, false);
+                Scope propagated = scopeManager.activate(span);
                 return propagated::close;
             };
         }
```
Ladicek, line 30: I'm looking at this again and I found this a bit weird. If the original thread had an active span, but not an active scope, the new thread will have the same active span, but it will _also_ have an active scope.

I have read through https://github.com/opentracing/opentracing-java/issues/267 and https://github.com/opentracing/opentracing-java/issues/291 and it seems to me that OpenTracing just actively fights against our progamming model :-)

But I guess there's nothing better that we could do.

PR #409, opened 2021-05-05
Java16 support

Ladicek:


Ladicek:
Wow, macOS does seem to have a rather unprecise scheduler :-)

michalszynkiewicz:
why do we have only windows checks? That's an interesting choice ;)

michalszynkiewicz (approved):

Ladicek:
Look at the latest commit, I'm doing some debugging as there's a weird failure that never happened on Linux. That commit will disappear and we'll run CI on 8, 11, 16 and 17-ea on Linux, Windows and Mac.

PR #427, opened 2021-05-20
Remove non-existent dependency from dependabot configuration

rhusar:
This configuration got carried over when dependabot suggested migration from dependabot-preview but it didn't account that this dependency has been removed completely.


Ladicek (approved):

Ladicek:
Indeed this only made sense on the 2.x branch and is now obsolete.

PR #443, opened 2021-06-21
more strict fallback method return type assignability check

Ladicek:


Ladicek:
This would be a breaking change, but the code changed here isn't present in any release yet. In 5.1.0 and prior, the branch that is changed here ended up in an infinite regress.

In other words, this PR adjusts #441 to behave more strictly. I believe that is correct in terms of the MP FT spec, because it says:

> The named fallback method must have the same return type as the annotated method (after resolving any type variables)

We have actually been more lenient, allowing fallback method's return type to be a _subtype_ of the original method's return type. This PR doesn't change that, but makes sure we don't add more leniency.

PR #452, opened 2021-07-02
@Blocking and @NonBlocking on classes

Ladicek:


implementation/fault-tolerance/src/main/java/io/smallrye/faulttolerance/CdiLogger.java
```diff
@@ -23,9 +23,13 @@
...
+    DefinitionException backoffAnnotationWithoutRetry(String backoffAnnotation, MethodDescriptor method);
 
-    @Message(id = 4, value = "Backoff annotation @%s present on class '%s', but @Retry is missing")
-    DefinitionException backoffOnClassWithoutRetry(String backoffAnnotation, Class<?> clazz);
+    DefinitionException backoffAnnotationWithoutRetry(String backoffAnnotation, Class<?> clazz);
+
+    @Message(id = 4, value = "Both @Blocking and @NonBlocking present on '%s'")
+    DefinitionException bothBlockingNonBlockingPresent(MethodDescriptor method);
+
+    DefinitionException bothBlockingNonBlockingPresent(Class<?> clazz);
```
Ladicek, line 34: This would be a breaking change, but messages with code 3 and above are not present in a release yet, so it's OK.

PR #479, opened 2021-08-16
Replaced Thread.sleep calls with Awaitility in CompletionStageThreadPoolBulkheadTest

hbelmiro:
Replaced Thread.sleep calls with Awaitility and small clean ups in CompletionStageThreadPoolBulkheadTest

### DIFF
--- a/implementation/core/src/test/java/io/smallrye/faulttolerance/core/bulkhead/CompletionStageThreadPoolBulkheadTest.java
+++ b/implementation/core/src/test/java/io/smallrye/faulttolerance/core/bulkhead/CompletionStageThreadPoolBulkheadTest.java
@@ -3,18 +3,13 @@ package io.smallrye.faulttolerance.core.bulkhead;
 import static java.util.concurrent.CompletableFuture.completedFuture;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
-import static org.assertj.core.api.Assertions.fail;
+import static org.awaitility.Awaitility.await;
 
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionStage;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.Semaphore;
-import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.TimeUnit;
+import java.util.Optional;
+import java.util.concurrent.*;
 
 import org.eclipse.microprofile.faulttolerance.exceptions.BulkheadException;
 import org.junit.jupiter.api.AfterEach;
@@ -132,7 +127,7 @@ public class CompletionStageThreadPoolBulkheadTest {
     }
 
     @Test
-    public void shouldLetMaxPlus1After1Failed() throws Exception {
+    public void shouldLetMaxPlus1After1Failed() {
         RuntimeException error = new RuntimeException("forced");
 
         Barrier delayBarrier = Barrier.noninterruptible();
@@ -170,7 +165,7 @@ public class CompletionStageThreadPoolBulkheadTest {
     }
 
     @Test
-    public void shouldLetMaxPlus1After1FailedCompletionStage() throws Exception {
+    public void shouldLetMaxPlus1After1FailedCompletionStage() {
         RuntimeException error = new RuntimeException("forced");
 
         Barrier delayBarrier = Barrier.noninterruptible();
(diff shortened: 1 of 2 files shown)


implementation/core/src/test/java/io/smallrye/faulttolerance/core/bulkhead/CompletionStageThreadPoolBulkheadTest.java
```diff
@@ -3,18 +3,13 @@
...
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionStage;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.Semaphore;
-import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.TimeUnit;
+import java.util.Optional;
+import java.util.concurrent.*;
```
Ladicek, line 12: Please don't.

implementation/core/src/test/java/io/smallrye/faulttolerance/core/bulkhead/CompletionStageThreadPoolBulkheadTest.java
```diff
@@ -226,42 +221,31 @@ public void shouldNotStartNextIfCSInProgress() throws Exception {
         party.organizer().waitForAll();
 
         CompletionStage<String> secondResult = bulkhead.apply(new InvocationContext<>(null));
-        waitUntilQueueSize(bulkhead, 1, 100);
+        waitUntilQueueSize(bulkhead, 1, 150);
```
Ladicek, line 224: I guess you have to increase the timeout to 150 millis because Awaitility's default "probe" interval is 100 millis? I'd rather change the interval in the `await()` call chain. The hand-written probe loop used 50 millis, I'd be fine with using 10 or 20 I think.

implementation/core/src/test/java/io/smallrye/faulttolerance/core/bulkhead/CompletionStageThreadPoolBulkheadTest.java
```diff
@@ -226,42 +221,31 @@ public void shouldNotStartNextIfCSInProgress() throws Exception {
...
-            if (bulkhead.getQueueSize() == size) {
-                return;
-            }
-        }
-        fail("queue not reached size " + size + " in " + timeout + " [ms]");
+    private static <V> void waitUntilQueueSize(CompletionStageThreadPoolBulkhead<V> bulkhead, int size, long timeout) {
+        await().atMost(Duration.ofMillis(timeout)).until(() -> bulkhead.getQueueSize() == size);
+    }
+
+    private static <V> CompletionStage<V> getSingleFinishedResult(List<CompletionStage<V>> results, long timeout) {
```
Ladicek, line 237: This split to 2 methods looks strange at first. It's probably fine, it just made me stop for a bit.

implementation/core/src/test/java/io/smallrye/faulttolerance/core/bulkhead/CompletionStageThreadPoolBulkheadTest.java
hbelmiro: Absolutely. My IDE was not configured properly.

implementation/core/src/test/java/io/smallrye/faulttolerance/core/bulkhead/CompletionStageThreadPoolBulkheadTest.java
hbelmiro: Actually, the hand-written probe used a timeout of 100 as argument and had a sleep(50) before each iteration. In the end, that could result in a timeout of 150. When I replaced with Awaitility and set a timeout of 100, a test started to fail. Then I increased to 150 and it passed.

Ladicek: Ah yea, you're right, it could be up to 150 millis originally. However, the current situation is basically equivalent to checking if the condition holds, and if not, waiting 100 millis and checking again. Perhaps we could just fold and unconditionally wait 100 millis and be done with it? :-)

implementation/core/src/test/java/io/smallrye/faulttolerance/core/bulkhead/CompletionStageThreadPoolBulkheadTest.java
hbelmiro: I preferred to split to increase readability. I found my initial non splitted version pretty polluted.

    private static <V> CompletionStage<V> getSingleFinishedResult(List<CompletionStage<V>> results, long timeout) {
        return await().atMost(Duration.ofMillis(timeout))
                      .<Optional<CompletionStage<V>>>until(() -> {
                          for (CompletionStage<V> result : results) {
                              if (result.toCompletableFuture().isDone()) {
                                  return Optional.of(result);
                              }
                          }
                          return Optional.empty();
                      }, Optional::isPresent)
                      .get();
    }

WDYT?

Ladicek: Yea it's probably fine. I didn't even know `until` can return a value :-)

implementation/core/src/test/java/io/smallrye/faulttolerance/core/bulkhead/CompletionStageThreadPoolBulkheadTest.java
hbelmiro: I'm not sure if I got it. Do you mean revert from 150 to 100?

implementation/core/src/test/java/io/smallrye/faulttolerance/core/bulkhead/CompletionStageThreadPoolBulkheadTest.java
hbelmiro: Me neither. I just found out :)

implementation/core/src/test/java/io/smallrye/faulttolerance/core/bulkhead/CompletionStageThreadPoolBulkheadTest.java
Ladicek: No, I mean current code is equivalent to

```java
if (bulkhead.getQueueSize() != size) {
    Thread.sleep(100);
    if (bulkhead.getQueueSize() != size) {
        throw new SomeException();
    }
}
```

Maybe we can just do that directly, instead of pretending there's a wait loop.

I guess my preference would be to make the Awaitility interval shorter, 10 or 20 millis:

```java
await().pollInterval(10, TimeUnit.MILLISECONDS).atMost(...)...
```

And 100 millis should be enough of a timeout IMHO.

hbelmiro: Yes, changing the pool interval to 10 made the test pass with 100 millis (including on my machine :-) )

Ladicek (approved):

Ladicek:
Thank you!

PR #480, opened 2021-08-18
Replace sleep calls with Awaitility in FutureThreadPoolBulkheadTest

hbelmiro:
We could also extract `waitUntilQueueSize()` and `getSingleFinishedThread()` from `FutureThreadPoolBulkheadTest` and `CompletionStageThreadPoolBulkheadTest` to new classes and reuse the same code in both test classes, since they are pretty similar. I don't think it's necessary, but if you do, just let me know.

### DIFF
--- a/implementation/core/pom.xml
+++ b/implementation/core/pom.xml
@@ -70,6 +70,12 @@
             <artifactId>assertj-core</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.awaitility</groupId>
+            <artifactId>awaitility</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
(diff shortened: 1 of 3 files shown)


Ladicek:
Could you please rebase on current `main`? Also, if you have multiple changes related to the same thing (such as this PR and #481), feel free to combine them into a single PR. I think that would be easier both for you and me :-)

hbelmiro:
> Could you please rebase on current `main`? Also, if you have multiple changes related to the same thing (such as this PR and #481), feel free to combine them into a single PR. I think that would be easier both for you and me :-)

Sure. And I did a big mess here in my branches :(
I'm going to close both PRs and create a new one.

Ladicek:
Sure, that works too!

PR #481, opened 2021-08-18
Replace `Thread#sleep` in `ScheduledExecutorTimeoutWatcherTest`

hbelmiro:


implementation/core/src/test/java/io/smallrye/faulttolerance/core/timeout/ScheduledExecutorTimeoutWatcherTest.java
```diff
@@ -74,21 +75,10 @@ public void notTimedOut() throws InterruptedException {
...
-                throw new RuntimeException("Interrupted waiting for " + message);
-            }
-            if (test.get()) {
-                return;
-            }
-        }
-        fail(message + " not satisfied in " + timeoutMs + "ms");
+        await(message).pollInterval(Duration.ofMillis(50))
+                .atMost(Duration.ofMillis(timeoutMs))
+                .until(test::get);
```
Ladicek, line 81: I think I'd just replace the calls to `assertThatWithin` with direct Awaitility calls, using `untilAsserted`. Something like

```java
await().pollInterval(Duration.ofMillis(10)).atMost(Duration.ofMillis(100)).untilAsserted(() -> {
    assertThat(watch.isRunning).as("watch not running").isFalse();
});
```

hbelmiro:
I'm closing this PR. I'll create a new one mixing this and https://github.com/smallrye/smallrye-fault-tolerance/pull/480.

PR #484, opened 2021-08-23
use TypeName instead of Class to represent types in the config system

Ladicek:


Ladicek:
Turns out this might not be necessary after all. I'll close this later today if that ends up true.

Ladicek:
Not necessary.

PR #576, opened 2022-02-10
Fix AsyncTypes to service load from the correct classloader instead o…

rhusar:
…f using the current thread context classloader

This is because you cannot guarantee the implementation is going to be available to the thread.
For instance, in the WildFly case, the implementation classes are not available in the deployment's classloader.
This manifests as:

Cannot deploy ftAsynchronous.war: {"WFLYCTL0062: Composite operation failed and was rolled back. Steps that failed:" => {"Operation step-1" => {"WFLYCTL0080: Failed services" => {"jboss.deployment.unit.\"ftAsynchronous.war\".WeldStartService" => "Failed to start service
    Caused by: org.jboss.weld.exceptions.DefinitionException: Exception List with 1 exceptions:
Exception 0 :
org.eclipse.microprofile.faulttolerance.exceptions.FaultToleranceDefinitionException: Invalid @Asynchronous on org.eclipse.microprofile.fault.tolerance.tck.asynchronous.AsyncClassLevelClient.serviceCS(java.util.concurrent.Future): must return java.util.concurrent.Future or


rhusar:
The only formatting problem I see is the line is "too long". Let me try breaking it. Otherwise I don't see a log of what's wrong.

Ladicek:
You're supposed to run `mvn clean compile` locally, the formatter will reformat things for you :-) (Though its output is sometimes truly despicable...)

Ladicek (approved):

Ladicek:
I'll do a 5.3.1 today.

rhusar:
@Ladicek Well I actually did that- but when committing I hated that extra space and though I did it inadvertently, so I reverted that change and opened a PR :-) I don't remember seeing a project that does this, but why not.

rhusar:
> I'll do a 5.3.1 today.

Sounds good, thanks.

rhusar:
@Ladicek While we are at it, do you need explicit PR to the v6 / jakarta branch?

PR #579, opened 2022-02-14
AsyncTypes class needs to service load in a privileged block to work with security manager enabled.

rhusar:


Ladicek (approved):

Ladicek:
I probably should have kept the `CompletionStage` type a special case :laughing:

Ladicek:
5.3.2 is in Central now.

PR #591, opened 2022-03-01
add when methods for allowing programmatic exception decisions

### LINKED ISSUE #590: allow programmatic exception decisions for circuit breaker, fallback and retry
Currently, the programmatic API only allows declarative exception decisions. That is, it is only possible to configure a set of exception types that are considered success and failure. An alternative approach should be added that consists of configuring an exception predicate. For example, instead of

```java
FaultTolerance.createCallable(this::action)
        .withFallback().handler(this::fallback).applyOn(RuntimeException.class).done()
        .build();
```

one should be able to write

```java
FaultTolerance.createCallable(this::action)
        .withFallback().handler(this::fallback).when(e -> e instanceof RuntimeException).done()
        .build();
```

Obviously the value of this approach is that the predicate may be more complex than a simple `instanceof` check.

Ladicek:
The `when` method is added to builders for circuit breaker, fallback
and retry and can't be combined with the existing methods that allow
declarative exception decisions (based on sets of exception types).

Resolves #590


vsevel:
I see this is part of `5.4.0`, and Quarkus 2.8.1 still references `5.3.1`. Will `5.4.0` be part of Quarkus `2.9.0`?

Ladicek:
I'll release 5.4.0 tomorrow, so yes.

Ladicek:
FYI, here's the PR to upgrade SmallRye Fault Tolerance in Quarkus: https://github.com/quarkusio/quarkus/pull/24674

PR #604, opened 2022-03-29
initial support for preconfigured, reusable fault tolerance

### LINKED ISSUE #589: ability to centralize fault tolerance configuration for multiple methods
Currently, with the declarative FT API, all methods must be annotated to be guarded with fault tolerance strategies:

```java
class MyService {
    @Fallback
    @Retry
    @CircuitBreaker
    void doSomething1() {
        ...
    }

    @Fallback
    @Retry
    @CircuitBreaker
    void doSomething2() {
        ...
    }
}
```

If all methods on one class are fine with using the same set of fault tolerance strategies, the annotations may be moved to the class:

```java
@Fallback
@Retry
@CircuitBreaker
class MyService {
    void doSomething1() {
        ...
    }

    void doSomething2() {
        ...
    }
}
```

But there is no other way to centralize fault tolerance configuration for multiple methods, especially if they are present on multiple classes. A typical example would be multiple RestClient interfaces, all pointing to the same external server. Those would probably like to share fault tolerance strategies completely.

The only way to centralize fault tolerance configuration right now is to use the programmatic API. There might be a way to leverage a programmatically created `FaultTolerance` object in a declarative way. Something like:

```java
@Produces
@Identifier("my-fault-tolerance")
static final FaultTolerance<Object> = ...;

...

@ApplyFaultTolerance("my-fault-tolerance")
public void doSomething() {
    ...
}
```

This particular approach has some issues:

- type safety
  - especially in case of fallback, where the user needs to return a value of the right type
- asynchrony
  - sync and async `FaultTolerance` objects are created differently (`FaultTolerance.create` vs. `FaultTolerance.createAsync` vs. `MutinyFaultTolerance.create`)
- ability to specify additional fault tolerance strategies right on the method (or its class)
  - easiest way would be to prevent that completely, but that's not exactly friendly

Need to think about this.

Ladicek:
Resolves #589.

### DIFF
--- a/api/src/main/java/io/smallrye/faulttolerance/api/ApplyFaultTolerance.java
+++ b/api/src/main/java/io/smallrye/faulttolerance/api/ApplyFaultTolerance.java
@@ -1,0 +1,93 @@
+package io.smallrye.faulttolerance.api;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import javax.interceptor.InterceptorBinding;
+
+import io.smallrye.common.annotation.Experimental;
+
+/**
+ * A special interceptor binding annotation to apply preconfigured fault tolerance.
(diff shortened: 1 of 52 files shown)


Ladicek:
Draft for now, because I still need to add documentation.

PR #621, opened 2022-04-11
get rid of JUnit Pioneer

Ladicek:
The `@SetSystemProprety` annotation/extension from JUnit Pioneer
was used on a couple places to change MP Config configuration.
In version 1.7.0, JUnit Pioneer makes a breaking change and
class-level annotations are now handled as method-level annotations
present on all methods, which leads to some tests failing
because MP Config configuration isn't changed early enough.

(An implicit assumption is made that class-level annotations
roughly correspond to the `{Before,After}AllCallback` extension
points of JUnit Jupiter, while method-level annotations roughly
correspond to the `{Before,After}EachCallback` extension points.
This assumption is, in my opinion, rather intuitive.)

In this commit, `@SetSystemProperty` from JUnit Pioneer is replaced
with our own `@WithSystemProperty` that behaves the original way.
Since this was the only usage of JUnit Pioneer, the dependency
on JUnit Pioneer is removed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,7 +60,6 @@
         <version.assertj>3.22.0</version.assertj>
         <version.awaitility>4.2.0</version.awaitility>
         <version.junit-jupiter>5.8.2</version.junit-jupiter>
-        <version.junit-pioneer>1.6.2</version.junit-pioneer>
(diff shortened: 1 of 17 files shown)


Ladicek:
Follows up on #620.

Issue #641, opened 2022-05-24
Unable to have logs on faulttolerance operations


manofthepeace:
I am using SR FT in a quarkus project. I'd like to be able to log what is going on where I have faulttolerance annotations.

Basic example, I would like to just basically log on retry, which api got retried, and maybe some context, if any. At the moment I do not know how to have that.

I tried few things without success.

1. I tried to put a decorator on some Ft bean, but ultimately I got nowhere with that. Or the decorator would not get called, or I would get a class not found exception on my own decorator itself (really dont understand that one..)
2. I tried observing CDI Events, as I saw the fireEvents in the retry class, but they dont seem to be cdi events, as I am not observing anything (or am I misusing it?)
3. Enabled `io.smallrye.faulttolerance.core.retry.Retry` to trace level, but it's too much and not enough at the same time, as I do not see the actual bean/beanmethod that got retried.

We have a lot of restClient in our project and use the `@Retry` a lot, but when things fails on the remotes, we cannot troubleshot at all at the moment, as we do not know if our app retried or not/how many times, or what have happened. So having the possibility to be able to have basic logs would be great, would be nice to manage the logs ourselves as well, via cdi events maybe, or a decorator.

Ladicek:
I'd generally suggest to put logs into the annotated method, as Fault Tolerance has almost no interesting application-specific information to log. In case of RestClient methods, that gets a little clunky -- you basically have to wrap the RestClient interface by a service class, put FT annotations on _that_ class, and route all invocations through it. But it's also the easiest thing possible.

Alternatively, it should be possible (though I don't recall how off the top of my head) to enable logging of the RestClient invocations themselves.

Also, fault tolerance emits metrics (though [Micrometer support](#265) is unfortunately still not there), which are pretty granular. To observe retries, that's probably the best way.

manofthepeace:
Thanks for your time and your reply.

I do understand that Fault Tolerance does not have much info of the application itself. In general I think just the class/methodName would be enough at least we know what calls where made by FT and in which case ( the ones in the RetryEvents class for example)

I already have a sort of service wrapping the RestClient, the retries are not on that service though, but I am unsure how to differentiate an actual call to the service, or a call from Fault Tolerance/retried. As our app can control via configuration the retries properties (including the abortOn/retryOn) it would really be useful to be able when we receive a log from outside seeing something like "ah ok FT retried twice in 10 sec and then that's why it failed with that message". In our current case we can only assume something happened.

[...]

PR #683, opened 2022-07-23
Add more information to Retry/Timeout logs

stephan-strate:
Extension for #641.

### DIFF
--- a/implementation/core/src/main/java/io/smallrye/faulttolerance/core/timeout/CompletionStageTimeout.java
+++ b/implementation/core/src/main/java/io/smallrye/faulttolerance/core/timeout/CompletionStageTimeout.java
@@ -34,7 +34,7 @@ public class CompletionStageTimeout<V> extends Timeout<CompletionStage<V>> {
         AtomicBoolean completedWithTimeout = new AtomicBoolean(false);
         Runnable onTimeout = () -> {
             if (completedWithTimeout.compareAndSet(false, true)) {
-                LOG.debug(description + " invocation timed out");
+                LOG.debug(description + " invocation timed out (" + timeoutInMillis + "ms)");
                 ctx.fireEvent(TimeoutEvents.Finished.TIMED_OUT);
                 result.completeExceptionally(timeoutException(description));
             }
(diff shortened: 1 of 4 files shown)


Ladicek (approved):

manofthepeace:
Funny coincidence, I did the almost same change last Friday locally. One thing I saw is that in the case where it is not the CompletionStage, the counter should be incremented by 1 in the log, since it starts at 0, so in the log it wont ever reach the maxAttempts.

For completionStage I had this;
`LOG.debugf("%s invocation failed, retrying (attempt %d of %d)", description, counter, maxRetries);`

for the non completionStage I had this;
`LOG.debugf("%s invocation failed, retrying (attempt %d of %d)", description, counter+1, maxRetries);`

otherswise if you have let say 3 retries , 2 of 3 will be the last which I think is counterintuitive. For the completion stage, it seems ok since it enters the if statement above the log when the counter is at 1, when it is actually doing the first retry.

So I think, in the current pr,  one counter will start at 1, and reach the maxRetries, the other will start at 0 and reach MaxRetries-1

what do you guys think? I can Pr it if it makes sense

Ladicek:
I admit I didn't try it, but I'm looking at the code and `counter == 0` in `Retry`, as well as `attempt == 0` in `CompletionStageRetry` is the initial invocation, so they seem consistent to me and no increment should be necessary...

manofthepeace:
I just ran it and it works as it should. Sorry for the noise.

implementation/core/src/main/java/io/smallrye/faulttolerance/core/timeout/CompletionStageTimeout.java
```diff
@@ -34,7 +34,7 @@ private CompletionStage<V> doApply(InvocationContext<CompletionStage<V>> ctx) {
         AtomicBoolean completedWithTimeout = new AtomicBoolean(false);
         Runnable onTimeout = () -> {
             if (completedWithTimeout.compareAndSet(false, true)) {
-                LOG.debug(description + " invocation timed out");
+                LOG.debug(description + " invocation timed out (" + timeoutInMillis + "ms)");
```
rhusar, line 37: BTW missing space before 'ms' - fixed in https://github.com/smallrye/smallrye-fault-tolerance/pull/685

rhusar:
This change adds a lot of unnecessary String concatenations – oped PR to fix https://github.com/smallrye/smallrye-fault-tolerance/pull/685

PR #685, opened 2022-07-25
Fix unnecessary String concatenation in debug/trace statements.

rhusar:


implementation/core/src/main/java/io/smallrye/faulttolerance/core/bulkhead/FutureThreadPoolBulkhead.java
```diff
@@ -60,8 +60,8 @@ private Future<V> doApply(InvocationContext<Future<V>> ctx) throws Exception {
                 }
 
                 if (LOG.isTraceEnabled()) {
-                    LOG.trace("Cancelling bulkhead task," + (event.interruptible ? "" : " NOT")
-                            + " interrupting executing thread");
+                    LOG.tracef("Cancelling bulkhead task, %s interrupting executing thread",
```
Ladicek, line 63: An extra space before `%s` :-)

```suggestion
                    LOG.tracef("Cancelling bulkhead task,%s interrupting executing thread",
```

Ladicek (approved):
I was meaning to do this myself (though not that aggressively), so thanks :-)

implementation/core/src/main/java/io/smallrye/faulttolerance/core/bulkhead/FutureThreadPoolBulkhead.java
rhusar: Ah... I see now. Fixed.

Ladicek (approved):

PR #721, opened 2022-10-10
Remove withType() references

jgallimore:
…ics to 5.0.0-M3. Remove withType() references, and update references to SharedMetricRegistries and LegacyMetricsExtension

### DIFF
--- a/implementation/fault-tolerance/src/main/java/io/smallrye/faulttolerance/metrics/MicroProfileMetricsProvider.java
+++ b/implementation/fault-tolerance/src/main/java/io/smallrye/faulttolerance/metrics/MicroProfileMetricsProvider.java
@@ -19,7 +19,6 @@ import jakarta.inject.Singleton;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.eclipse.microprofile.metrics.Metadata;
 import org.eclipse.microprofile.metrics.MetricRegistry;
-import org.eclipse.microprofile.metrics.MetricType;
 import org.eclipse.microprofile.metrics.MetricUnits;
 import org.eclipse.microprofile.metrics.Tag;
 import org.eclipse.microprofile.metrics.annotation.RegistryType;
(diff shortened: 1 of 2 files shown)


jgallimore:
Thanks for the feedback. I hope the PR is some help, even if it isn't merged in. From my own perspective, I'm attempting to get a milestone build of TomEE with Metrics 5 and FT 4.0.2.

jgallimore:
Also, I'm happy to do rework on this with feedback from yourselves if that helps. :)

Ladicek:
No, this absolutely helps, I'll submit some PRs and this will help me verify.

Ladicek:
So the `jakarta` branch is currently on Jakarta EE 9 / MicroProfile 5. At some point, we'll want to switch to Jakarta EE 10 / MicroProfile 6. Perhaps even first move to Jakarta EE 10 and stay with MicroProfile 5, and then move to MicroProfile 6. In the SmallRye Fault Tolerance 6.0.0 announcement, I wrote that SmallRye Fault Tolerance will move to Jakarta EE 10 in the 6.x stream -- I don't think this requires a new major, but I can be convinced otherwise.

From MP Metrics perspective, we should be able to build SmallRye Fault Tolerance against MP Metrics 4 and run against both MP Metrics 4 and 5. (In fact, that should be the current state. To release from the `jakarta` branch, I always release from `main` first, then merge the tag into `jakarta`, and release that. If you replicate that process locally, with current state of `main` and `jakarta`, the outcome should be runtime compatible with MP Metrics 5.) So this shouldn't require a new major either.

Aside: I had no idea WildFly also moves to Micrometer. There's an integration in SmallRye Fault Tolerance, but it's not exposed through the CDI extension. Feel free to get in touch offline if you want to explore that.

rhusar:
> So the `jakarta` branch is currently on Jakarta EE 9 / MicroProfile 5. At some point, we'll want to switch to Jakarta EE 10 / MicroProfile 6. Perhaps even first move to Jakarta EE 10 and stay with MicroProfile 5, and then move to MicroProfile 6. In the SmallRye Fault Tolerance 6.0.0 announcement, I wrote that SmallRye Fault Tolerance will move to Jakarta EE 10 in the 6.x stream -- I don't think this requires a new major, but I can be convinced otherwise.

I don't suspect any changes are needed going EE 9 to 10, but I haven't looked closely.

> From MP Metrics perspective, we should be able to build SmallRye Fault Tolerance against MP Metrics 4 and run against both MP Metrics 4 and 5. (In fact, that should be the current state. To release from the `jakarta` branch, I always release from `main` first, then merge the tag into `jakarta`, and release that. If you replicate that process locally, with current state of `main` and `jakarta`, the outcome should be runtime compatible with MP Metrics 5.) So this shouldn't require a new major either.

I think that sounds good.

> Aside: I had no idea WildFly also moves to Micrometer. There's an integration in SmallRye Fault Tolerance, but it's not exposed through the CDI extension. Feel free to get in touch offline if you want to explore that.

Well, I wouldn't call it 'move', that would be premature as it's only added to the WildFly Preview distribution, not the default one (https://issues.redhat.com/browse/WFLY-14947).

jgallimore:
@Ladicek @rhusar I've stripped this down to just the removal of the withType() calls. That seems to pass the TCK in TomEE with MP Metrics 5.0.

Ladicek:
I've merged that to main already: https://github.com/smallrye/smallrye-fault-tolerance/pull/725 so that will be in 5.6.0 and also 6.1.0.

PR #727, opened 2022-10-13
CI: Drop no longer last JDK GA (18) + move JDK 19 to GA + add next JD…

rhusar:
…K 20 EA


Ladicek:
Ah ah, I wanted to do this a few days ago and totally forgot about it. Thanks!

Ladicek (approved):

PR #789, opened 2023-03-02
Allow FaultToleranceExtension to switch metrics integration provider.

rhusar:


rhusar:
@Ladicek Please have a look, perhaps we can do something smarter/more useful, but this would suffice for WF integration.

Ladicek:
I think we'll eventually have to provide a way for the extension to discover the expected metrics integration in some other way, for the cases when CDI container instantiates the extension on its own, but this is good enough when the integrator can instantiate the extension themselves.

PR #807, opened 2023-04-06
fix lookup of FaultToleranceSpi when multiple implementations are present

### LINKED ISSUE #806: FaultToleranceSpiAccess best candidate lookup issue
The FaultToleranceSpiAccess instantiateSpi function relies on ServiceLoader to load all the FaultToleranceSpi candidates. It has a logic to compare priority in order to find the FaultToleranceSpi best candidate in order to be use. The default priority has been defined https://github.com/smallrye/smallrye-fault-tolerance/blob/main/api/src/main/java/io/smallrye/faulttolerance/api/FaultToleranceSpiAccess.java#L23 and compare logic has been defined https://github.com/smallrye/smallrye-fault-tolerance/blob/main/api/src/main/java/io/smallrye/faulttolerance/api/FaultToleranceSpiAccess.java#L28. However bestCandidatePriority has not been changed and remained as `Integer.MIN_VALUE` when the best candidate has been switched `bestCandidate = candidate;` which leads to the result of the next candidate will always replace the previous one as best candidate. I am not sure whether this is intended or not but feels more like a bug.

Ladicek:
Fixes #806

### DIFF
--- a/api/src/main/java/io/smallrye/faulttolerance/api/FaultToleranceSpiAccess.java
+++ b/api/src/main/java/io/smallrye/faulttolerance/api/FaultToleranceSpiAccess.java
@@ -27,6 +27,7 @@ public final class FaultToleranceSpiAccess {
             }
             if (candidate.priority() > bestCandidatePriority) {
                 bestCandidate = candidate;
+                bestCandidatePriority = candidate.priority();
             }
         }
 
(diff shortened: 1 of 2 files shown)


cl8au (approved):
LGTM, thanks for helping us fix the issue :)

Issue #908, opened 2023-11-03
Programmatic Asynchronous API is difficult to understand and verbose


magicprinc:
```CompletionStage<ResultType> FaultTolerance.call( Callable<CompletionStage<ResultType> )``` is a party killer.

Shouldn't it be ```CompletionStage<ResultType> FaultTolerance.call( Callable<ResultType> )```?

It is not obvious, how does it actually work and what does it protect...

**Comment by Ladicek** [comment]
It shouldn't, no. That would only be possible for blocking, thread offload based concurrency, while we also need to support non-blocking, event loop based concurrency. If you only use thread offloads, you can create a simple helper function:

```java
public static <T> CompletionStage<T> wrap(Callable<T> task) {
    try {
        return CompletableFuture.completedFuture(task.call());
    } catch (Exception e) {
        return CompletableFuture.failedFuture(e);
    }
}
```

magicprinc:
It makes sense. 👍

Another unclear moment:

```.withThreadOffload(true)```
+
```.call(()->CompletableFuture.supplyAsync(() -> externalService.hello())```


Does `withThreadOffload(true)` mean that `Callable.call` (== CompletableFuture.supplyAsync) will run in another thread?

CompletableFuture.supplyAsync is a very light operation.
In this case, it makes not much sense.
Could you show more practical example?

Ladicek:
`withThreadOffload(true)` makes sure the guarded operation is executed on an extra thread (the thread pool is not currently configurable, but that's the purpose of #905). If you configure `withThreadOffload(true)`, you don't need to do additional thread offload by `CompletableFuture.supplyAsync()`.

Your code above basically does two thread offloads in a row.

Issue #841, opened 2023-06-12
Integrate programmatic API with metrics


ranjanashish:
Hi, I am using SmallRye Fault Tolerance in a Quarkus project. This is a request to integrate programmatic API of SmallRye Fault Tolerance with metrics.

Ladicek:
Understood. This will happen eventually, but is not high on my priority list.

magicprinc:
See also https://github.com/smallrye/smallrye-fault-tolerance/issues/919

+ grafana_dashboard.json would be great :-)

Ladicek:
I have an example Grafana dashboard for the fault tolerance metrics here: https://github.com/Ladicek/fault-tolerance-demo It is heavily based on the OpenLiberty MP Fault Tolerance Grafana dashboard, so kudos to them!

PR #910, opened 2023-11-03
Minor improvements

magicprinc:
1. FaultTolerance.complete as in https://github.com/smallrye/smallrye-fault-tolerance/issues/908
2. Collections.newSetFromMap(new ConcurrentHashMap<>()) → ConcurrentHashMap.newKeySet()
3. CompletionStages 3 lines to 1 (CompletableFuture.failedFuture)
4. .editorconfig
5. support of https://jitpack.io/ (It works: https://jitpack.io/#magicprinc/smallrye-fault-tolerance/6.2.6 )

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+end_of_line = lf
+insert_final_newline = true
--- a/implementation/core/src/main/java/io/smallrye/faulttolerance/core/apiimpl/BasicCircuitBreakerMaintenanceImpl.java
+++ b/implementation/core/src/main/java/io/smallrye/faulttolerance/core/apiimpl/BasicCircuitBreakerMaintenanceImpl.java
@@ -1,7 +1,6 @@
 package io.smallrye.faulttolerance.core.apiimpl;
 
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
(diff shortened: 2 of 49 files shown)


magicprinc:
@Ladicek I hope you find it useful. The last thing I have in my mind is: https://github.com/smallrye/smallrye-fault-tolerance/issues/841
No more complaints: your project is the best 👍

Ladicek:
EditorConfig is good. `ConcurrentHashMap.newKeySet()` is good. The missing `@Override` annotations is good.

The `CompletionStages` methods are a remnant of Java 8 times, the 2 methods just need to be removed and the corresponding `CompletableFuture` methods should be used directly.

I don't understand the need for special jitpack configuration, per the documentation at https://jitpack.io/, it should work out of the box. Can you please explain?

I will not accept the helper methods on `FaultTolerance`. That is something you should maintain on your own.

Thanks for the PR! I took the liberty of force-pushing and will merge once CI gets green.

Ladicek (approved):

magicprinc:
@@Ladicek About jetpack.io: interesting! I am user of Gradle, so: Either jetpack supports Maven better or they have upgraded default JDK version recently:
With Gradle projects, I have always been receiving compilation errors because of default JDK 8.
But this project without extra configs shows: https://jitpack.io/com/github/smallrye/smallrye-fault-tolerance/6.2.6/build.log
```
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF-8 -Dhttps.protocols=TLSv1.2
Apache Maven 3.6.1 (d66c9c0b3152b2e69ee9bac180bb8fcc8e6af555; 2019-04-04T19:00:29Z)
Maven home: /opt/apache-maven-3.6.1
Java version: 11.0.2, vendor: Oracle Corporation, runtime: /usr/lib/jvm/jdk-11
Default locale: en, platform encoding: UTF-8
OS name: "linux", version: "4.10.0-28-generic", arch: "amd64", family: "unix"
```

So you either can drop jitpack.yml or keep it to have precise control of JDK version. Sorry 🤷‍♀️

Ladicek:
No worries. The [documentation](https://docs.jitpack.io/building/#java-version) says that

> Maven projects that specify a target version in their pom will be built with that target version.

So I don't think we need a `jitpack.yml` at all. So I just dropped it.

magicprinc:
I add it here

Class
```io.smallrye.faulttolerance.core.timeout.ScheduledExecutorTimeoutWatcher```
is not used, probably replaced by TimerTimeoutWatcher

Ladicek:
That is correct. `ScheduledExecutorTimeoutWatcher` was the original implementation, before I implemented the timer. I left it in place in case, for no particular reason, we could as well remove it, because the timer doesn't seem to have caused any issues.

PR #922, opened 2023-11-18
InvocationContext micro optimizations

magicprinc:
1) in tests data.size() is never > 2: set initialCapacity from 16 to 4 
2) don't do extra work for absent eventHandler


Ladicek (approved):

Ladicek:
Makes sense.

PR #923, opened 2023-11-18
Concept: delete deprecated scheduler

magicprinc:
Delete deprecated (unused) ScheduledExecutorTimeoutWatcher

Delete extra wrapper TimeoutWatch: use TimerTask directly, spare ~ 20 bytes per call

-2 Classes :-)


Ladicek:
I agree with deleting the `ScheduledExecutorService`-based timer implementation, but I don't feel good about breaking the `TimeoutWatch` abstraction.

Ladicek:
We could possibly remove the entire `TimeoutWatcher` abstraction altogether, it probably isn't all that useful.

Ladicek:
Superseded by #927.

PR #924, opened 2023-11-18
Concept: Access to InvocationContext from Callable

magicprinc:
InvocationContext: API to access InvocationContext from Callable
to store data between calls in session context

### DIFF
--- a/implementation/core/src/main/java/io/smallrye/faulttolerance/core/InvocationContext.java
+++ b/implementation/core/src/main/java/io/smallrye/faulttolerance/core/InvocationContext.java
@@ -14,9 +14,13 @@ public final class InvocationContext<V> implements Callable<V> {
         this.delegate = delegate;
     }
 
-    @Override
+    @Override @SuppressWarnings("unchecked")
     public V call() throws Exception {
-        return delegate.call();
+        if (delegate instanceof FaultToleranceStrategy){ // API to access InvocationContext from Callable (store data in session context)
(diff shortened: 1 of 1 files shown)


Ladicek:
`FaultToleranceStrategy` should not be treated as public API.

magicprinc:
BTW, Strictly speaking, It is not treated as public: 
It is called from InvocationContext - inside same package :-)

But anyway, it was just a nice concept: ad hoc solution for custom FaultToleranceStrategies in pipeline

Ladicek:
> custom FaultToleranceStrategies in pipeline

is exactly what I had in mind when I said that "`FaultToleranceStrategy` should not be treated as public API" :-) All the APIs in `io.smallrye.faulttolerance.core` were designed as an implementation detail, so there's little care taken about backward compatibility. I actually have an internal refactoring in mind that might end up changing `FaultToleranceStrategy` in an incompatible manner. Not guaranteed that it will happen, but it's possible -- and if it does happen, that doesn't necessarily require a major version bump or release note or anything, because it's an internal API.

Also, all fault tolerance strategies in the chain just delegate directly to the "next" fault tolerance strategy, and there should only be _one_ that calls `InvocationContext.call()`. So it's really about properly constructing that chain.

magicprinc:
I have had to change my mindset to migrate from https://failsafe.dev/ 
and I still have "phantom pains".
But after all, I miss only metrics now :-)
And new release with all these last changes :-)

I wish your project great success!

PR #925, opened 2023-11-18
FaultToleranceImpl.BuilderImpl.withThreadOffload: allow disabling in sync mode

magicprinc:
FaultToleranceImpl.BuilderImpl.withThreadOffload

withThreadOffload(false) must be allowed in any mode (sync or async)


Ladicek:
Makes sense.

Ladicek (approved):

PR #929, opened 2023-11-20
prevent arithmetic overflow and reduce allocations in time conversions

magicprinc:
See java.lang.Math#multiplyExact(long, long)

ArithmeticException can be thrown

and temporary Duration objects are created

### DIFF
--- a/implementation/core/src/main/java/io/smallrye/faulttolerance/core/apiimpl/FaultToleranceImpl.java
+++ b/implementation/core/src/main/java/io/smallrye/faulttolerance/core/apiimpl/FaultToleranceImpl.java
@@ -1,8 +1,8 @@
 package io.smallrye.faulttolerance.core.apiimpl;
 
 import static io.smallrye.faulttolerance.core.Invocation.invocation;
+import static io.smallrye.faulttolerance.core.util.Durations.timeInMillis;
 
-import java.time.Duration;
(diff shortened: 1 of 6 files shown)


magicprinc:
There are two alternatives (and two commits)
1) we create Duration (new temporary object) and catch ArithmeticException
or
2) we use TimeUnit static function (no object creation, no exceptions), but it supports only NANOS..DAYS (no WEEKS, MONTHS, YEARS). I think for fault-tolerance library such units are not relevant…

Ladicek:
Good point, I didn't realize that. I generally think that overflow would be rare, so we could just wrap the `Duration.toMillis()` call in a `try`/`catch` and return `Long.MAX_VALUE`, but if we can avoid allocating, that's even better :-)

Ladicek (approved):

Ladicek:
I removed the `Thread.yield()` call, because I haven't seen conclusive evidence that this is actually helpful. Do you have any benchmarks that motivated this change?

Ladicek:
`Thread.onSpinWait()` would be a mistake here, the `ThreadSleepDelay` is not busy waiting / spinning.

If I understand correctly, you attempted to play nice in the cooperative multi-threading world of virtual threads, but I believe the retry strategy is not the right place for that. If the guarded logic is heavy on computations and monopolizes the carrier thread, it is _its_ job to play nice.

Issue #987, opened 2024-03-23
Semaphore behavior in Bulkhead strategy (add queueSize and timeout)


magicprinc:
With JDK21 we have Virtual Threads: they are cheap, and you can block them without any thoughts

Could you add an option to Semaphore based Bulkhead strategies (e.g. timeout: long millis),
which, when enabled, changes the behavior of Semaphore from
`tryAcquire()`
to 
`tryAcquire(long timeout, TimeUnit unit/*MILLISECONDS*/)`

So, if one runs a code with Bulkhead and sets limit to 10, then only 10 threads would run concurrently and other (millions) would be put into "waiting queue"

The application code itself would be much simpler.

Now I have to use plain Semaphore in addition to FT :-(

magicprinc:
Here is an example, 
```
@Test
void _bulkhead () {
  // the other side has set a contractual limit to 2 concurrent connections
  var ft = FaultTolerance.<Integer>create().withDescription("_bulkhead")
    .withBulkhead().limit(2).done()
    .build();

  var total = new AtomicInteger();
  var concurrent = new AtomicInteger();

  // we load tasks (the queue length is controlled) and run them in Virtual Thread
  for (int i=0; i < 20; i++){
    Runnable task = ()->{
      try {
        total.incrementAndGet();
        ft.call(()->{
          concurrent.incrementAndGet();
          try { Thread.sleep(20); } catch (Throwable ignore){}
          System.out.printf("Data sent: total=%s, concurrent=%s%n", total, concurrent);
          concurrent.decrementAndGet();
          return null;
        });
      } catch (Exception e) {
        System.out.printf("FT.call failed: total=%s%n", total);
      }
    };

    JThread.VT.execute(task);// run in VT using VT executor
  }
}
```

here is the result
```
FT.call failed: total=16
FT.call failed: total=18
FT.call failed: total=19
FT.call failed: total=20
FT.call failed: total=20
FT.call failed: total=20
FT.call failed: total=20
FT.call failed: total=20
FT.call failed: total=20
FT.call failed: total=20
FT.call failed: total=20
FT.call failed: total=20
FT.call failed: total=20
FT.call failed: total=20
FT.call failed: total=20
FT.call failed: total=20
FT.call failed: total=20
FT.call failed: total=20
Data sent: total=20, concurrent=2
Data sent: total=20, concurrent=1
```

magicprinc:
[...]

PR #1015, opened 2024-04-26
Sync/Blocking Bulkhead implementation with virtual Queue

magicprinc:
TwoSemaphoreBulkhead

https://github.com/smallrye/smallrye-fault-tolerance/issues/987


magicprinc:
Now with unit tests, including example use case!

Ideal for virtual threads

PR #1034, opened 2024-07-12
make LazyFaultTolerance lock-free

mariofusco:
This pull request is intended to fix the issue reported here https://github.com/quarkusio/quarkus/issues/41314 

Its main drawback is in the fact that with this change multiple threads could concurrently call two or more times the `FaultTolerance builder`. I don't know how heavy that invocation actually is, but in case you think that this is not acceptable, the other alternative that I see is replacing that synchronized block with a `ReentrantLock`. Please advice on which of the 2 solutions you prefer.

### DIFF
--- a/implementation/core/src/main/java/io/smallrye/faulttolerance/core/apiimpl/LazyFaultTolerance.java
+++ b/implementation/core/src/main/java/io/smallrye/faulttolerance/core/apiimpl/LazyFaultTolerance.java
@@ -1,6 +1,7 @@
 package io.smallrye.faulttolerance.core.apiimpl;
 
 import java.util.concurrent.Callable;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
 
 import io.smallrye.faulttolerance.api.FaultTolerance;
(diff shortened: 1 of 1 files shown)


implementation/core/src/main/java/io/smallrye/faulttolerance/core/apiimpl/LazyFaultTolerance.java
```diff
@@ -36,15 +37,13 @@ public void run(Runnable action) {
     }
 
     private FaultTolerance<T> instance() {
-        FaultTolerance<T> instance = this.instance;
+        FaultTolerance<T> instance = this.instance.get();
```
franz1981, line 40: this is making `instance::get` to be called more than once, changing the expected behaviour for users.
Using ReentrantLock(s) is preferrable here - or use a 2 state lock-free behaviour instead

franz1981 (changes requested):

Ladicek:
The problem isn't in how heavy the FT builder is, it's in how it creates certain resources; notably, the registration with the circuit breaker maintenance must be done exactly once (subsequent invocations end up with an error IIRC). So a `ReentrantLock` would be preferred.

Thanks for working on this BTW, I totally missed that Quarkus issue!

mariofusco:
Replace by https://github.com/smallrye/smallrye-fault-tolerance/pull/1035

PR #1036, opened 2024-07-16
add @BeforeRetry

### LINKED ISSUE #259: prototype "before retry action" support

Ladicek:
Fixes #259


Ladicek:
This is a draft, because tests are missing. Documentation probably also needs some improvement.

Publishing early to solicit feedback from @cescoffier and @geoand. I don't think you need to review everything, just look at the `api` module (especially the `@BeforeRetry` annotation). I'm especially interested in knowing whether these 2 aspects work for you:

- the before retry method is always `void` and has no parameters
- it must not block, it must be short and it must not throw

It would be possible to add parameters, similarly to how `@Fallback` works. It would probably also be possible to make it return `CompletionStage<Void>` (and other async types) so that async work is possible (in addition to `void` if async work is not necessary).

That's probably all.

geoand:
Nice!

I'll have a look soon

cescoffier:
LGTM,

We will likely automatically inject that annotation in Quarkus langchain4j.

geoand:
Lovely, we will certainly make use of it!

Ladicek:
OK, I take it the current API is good for you. I'll merge this and release. If it turns out the API is not sufficient, we can come back to it.

PR #1045, opened 2024-08-02
fix method search in SecurityActions.findMethod()

Ladicek:
The previous implementation considered all superclasses, but only implemented interfaces of the bean class. The new implementation also considers implemented interfaces of all superclasses and superinterfaces.

### DIFF
--- a/implementation/fault-tolerance/src/main/java/io/smallrye/faulttolerance/internal/SecurityActions.java
+++ b/implementation/fault-tolerance/src/main/java/io/smallrye/faulttolerance/internal/SecurityActions.java
@@ -27,8 +27,10 @@ import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.security.PrivilegedActionException;
 import java.security.PrivilegedExceptionAction;
+import java.util.ArrayDeque;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Deque;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


implementation/fault-tolerance/src/main/java/io/smallrye/faulttolerance/internal/SecurityActions.java
```diff
@@ -189,18 +196,18 @@ private static Set<Method> findMethod(Class<?> beanClass, Class<?> declaringClas
...
-            clazz = clazz.getSuperclass();
-        }
 
-        for (Class<?> iface : declaringClass.getInterfaces()) {
-            for (Method m : iface.getMethods()) {
-                result.add(m.getName());
+            if (clazz.getSuperclass() != null) {
+                worklist.add(clazz.getSuperclass());
             }
+            Collections.addAll(worklist, clazz.getInterfaces());
```
zakkak, line 210: Are these really needed?

The comment above says:

```
// if we find a matching method on the bean class or one of its superclasses or superinterfaces,
// then we have to check that the method is either identical to or an override of a method that:
// - is declared on a class which is a superclass of the declaring class, or
// - is declared on an interface which implemented by the declaring class
```

which according to my understanding (and assuming it's correct) means that we only need to get the methods declared on the declaring class and its superclasses, plus the methods declared on the interfaces implemented by the declaring class only.

I believe the previous version of the code was fine and perhaps it would also suffice to replace `iface.getMethods()` with `iface.getDeclaredMethod()` (so that we don't need the whole hierarchy).

zakkak (changes requested):
Since SmallRye is looking for a specific method with specific parameters I think it would be better to use `Class.getMethod(...)` on each class/interface instead of getting all methods and then filtering them by name.

Ladicek:
> Since SmallRye is looking for a specific method with specific parameters I think it would be better to use `Class.getMethod(...)` on each class/interface instead of getting all methods and then filtering them by name.

Yeah, that unfortunately only works for `public` methods. And `getDeclaredMethod()` doesn't return inherited methods. That's why we need to traverse the inheritance hierarchy manually.

zakkak:
> > Since SmallRye is looking for a specific method with specific parameters I think it would be better to use `Class.getMethod(...)` on each class/interface instead of getting all methods and then filtering them by name.
> 
> Yeah, that unfortunately only works for `public` methods. And `getDeclaredMethod()` doesn't return inherited methods. That's why we need to traverse the inheritance hierarchy manually.

Ooops, right. What about `Class.getDeclaredMethod(...)`?

PR #1048, opened 2024-09-19
make ThreadTimer not singleton again

### LINKED ISSUE #1047: Timer should not be a singleton
In WildFly, if there are multiple deployments using SmallRye Fault Tolerance, multiple timers are created. This used to work fine, but fails now. We need to rollback the singleton nature of `ThreadTimer`, while preserving metrics. The only possible way I can think of is adding a tag (or attribute) to the `ft.timer.scheduled` metric identifying the timer.

Ladicek:
This is for multi-deployment environments, like WildFly, where multiple applications may exist and each will have its own timer.

Fixes #1047

### DIFF
--- a/doc/modules/ROOT/pages/reference/metrics.adoc
+++ b/doc/modules/ROOT/pages/reference/metrics.adoc
@@ -45,7 +45,8 @@ The behavior of the timer thread can be observed through the following metrics:
 | Type | `Gauge<Integer>`
 | Unit | None
 | Description | The number of tasks that are currently scheduled (for future execution) on the timer.
-| Tags | None
+| Tags
+a| * `id` - the ID of the timer, to distinguish multiple timers in a multi-application environment
 |===
 
 == Micrometer Support
--- a/implementation/core/src/main/java/io/smallrye/faulttolerance/core/timer/ThreadTimer.java
+++ b/implementation/core/src/main/java/io/smallrye/faulttolerance/core/timer/ThreadTimer.java
@@ -10,6 +10,7 @@ import java.util.concurrent.ConcurrentSkipListSet;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.LockSupport;
 
 import io.smallrye.faulttolerance.core.util.RunnableWrapper;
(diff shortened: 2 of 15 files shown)


Ladicek:
@rhusar Here's the fix for singleton `Timer`. If you could test this with WildFly, that would be great!

rhusar:
> @rhusar Here's the fix for singleton `Timer`. If you could test this with WildFly, that would be great!

Thanks, I ll have a look and report back.

Ladicek:
Hi @rhusar, any news?

rhusar:
@Ladicek TBH having a bit of a problem reproducing this, can you elaborate the conditions that are necessary to trigger this? Basically I have 2 concurrent deployment both using a Retry, but that doesn't trigger the problem.

Ladicek:
Humm, OK, that's weird. Did you try to _call_ the `@Retry` method in both apps? There's certain level of laziness.

rhusar:
> Humm, OK, that's weird. Did you try to _call_ the `@Retry` method in both apps? There's certain level of laziness.

Yes. I am starting to think there is another condition required to trigger the Lazy resolution bit.. Also, in the user report this appears to not require calling the methods since this happens in `WeldStartService`.

Ladicek:
Ah right, good point. The timer is created eagerly. Well, then I don't know what's the issue...

mmusenbr:
We have bound the Retry to a reactive-Messaging channel, I can try to strip that down to a sharable example, but I do not have the time to do that today (CEST timezone).

rhusar:
> Orignal Issue-Reporter in zulip here: We have bound the Retry to a reactive-Messaging channel, I can try to strip that down to a sharable example, but I do not have the time to do that today (CEST timezone).

Thanks @mmusenbr, that would be immensely helpful (the reactive messaging subsystem was added after integrating MP Fault Tolerance so I guess that's why we don't have this covered in WildFly test suite better..)

Ladicek:
With this PR, the exception no longer occurs. Looks good enough to me.

rhusar:
> With this PR, the exception no longer occurs. Looks good enough to me.

Sounds good. I was packaging these in an EAR so perhaps that has an effect too.

Ladicek:
Oh I have no idea how that works in an EAR. Two separately deployed WARs certainly trigger the problem.

rhusar:
@Ladicek Just rewritten the test to use separate top-level deployments = they fail and pass with the fix. So clearly, there is some context sharing going on in an EAR that doesn't trigger this.

Excellent, now we will just wait for 6.4.1 whenever we can get it ;-)

FYI @mmusenbr no need for the reproducer, thx!

fabiobrz:
> @Ladicek Just rewritten the test to use separate top-level deployments = they fail and pass with the fix. So clearly, there is some context sharing going on in an EAR that doesn't trigger this.
> 
> Excellent, now we will just wait for 6.4.1 whenever we can get it ;-)
> 
> FYI @mmusenbr no need for the reproducer, thx!

Sorry to be joining the party too late. We do have a failing test which is a good reproducer, here: https://github.com/jboss-eap-qe/eap-microprofile-test-suite/blob/master/microprofile-fault-tolerance/src/test/java/org/jboss/eap/qe/microprofile/fault/tolerance/UndeployDeployTest.java#L125

fabiobrz:
> If only it ran on CI ;-)

Eh... we're working on it :smiling_face_with_tear: 

> Yes, the issues are:

Thanks!

PR #1057, opened 2024-10-08
static resolution of FT methods (@Fallback.fallbackMethod, @BeforeRetry.methodName)

### LINKED ISSUE #1055: static resolution of fallback / before-retry methods
Currently, fallback / before-retry methods (`@Fallback.fallbackMethod`, `@BeforeRetry.methodName`) are resolved dynamically, at the same time the other configuration is resolved. This is too late for native image compilation in Quarkus -- we need to know the full reflection configuration during build.

Therefore, this configuration must be resolved "statically". For the CDI portable extension (fully runtime behavior), this means during fault tolerance method discovery. For Quarkus, this means in the Quarkus extension. This will require porting the reflective code to discover methods to Jandex, which is something we'll just have to live with.

Ladicek:
Fixes #1055


Ladicek:
This is the SmallRye Fault Tolerance part. There will also be a Quarkus counterpart to this.

PR #1059, opened 2024-10-15
ITestListener cleanup: remove redundant identical default method impl…

rhusar:
…ementations

Random cleanup.


Ladicek (approved):
Thanks!

PR #1062, opened 2024-10-22
add FaultTolerance.BulkheadBuilder.enableVirtualThreadsQueueing()

### LINKED ISSUE #987: Semaphore behavior in Bulkhead strategy (add queueSize and timeout)

Ladicek:
This method enables bulkhead queueing for synchronous actions executed on virtual threads. The implementation is a simple two-semaphores bulkhead, which has in fact already existed as `FutureThreadPoolBulkhead`. This commit generalizes that class and leaves `FutureThreadPoolBulkhead` as a special case, nearly empty, for guarding asynchronous actions returning `Future<V>`.

Fixes #987
Closes #1015

### DIFF
--- a/api/src/main/java/io/smallrye/faulttolerance/api/FaultTolerance.java
+++ b/api/src/main/java/io/smallrye/faulttolerance/api/FaultTolerance.java
@@ -410,6 +410,18 @@ public interface FaultTolerance<T> {
             BulkheadBuilder<T, R> queueSize(int value);
 
             /**
+             * Enables bulkhead queueing for synchronous actions executed on virtual threads.
+             * This makes it possible to call {@link #queueSize(int)} even if this builder does
+             * not configure fault tolerance for asynchronous actions.
+             * <p>
+             * If you use this method, you <strong>have to ensure</strong> that the guard
+             * is executed on a virtual thread.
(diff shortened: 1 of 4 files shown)


magicprinc:
I love it!

Only a tiny question of optimization:

Isn't `AtomicReference<Thread> executingThread = new AtomicReference<>(Thread.currentThread());` an overkill?

You only read it, so it can be 
`final Thread executingThread = Thread.currentThread()`?

Ladicek:
Yeah, that's possible. I might have had some mutations there in the past, who knows :shrug: Removing the `AtomicReference` should be possible.

PR #1069, opened 2024-10-25
Make metric providers properly lazy

Ladicek:
This commit turns injection of metric registries in metric providers into a lazy variant (`Provider<>`), and turns field injection into constructor injection. This is all to make `CompoundMetricsProvider` work properly in case some of the metric providers actually cannot function properly, because the respective dependencies do not exist.


Ladicek:
This should fix the issue found in https://github.com/smallrye/smallrye-reactive-messaging/pull/2801

PR #1079, opened 2024-11-19
add ability to disable individual metric providers

### LINKED ISSUE #1078: Exposing MP Config properties for individually disabling metrics providers
Following up on fixing CompoundMetricsProvider https://github.com/smallrye/smallrye-fault-tolerance/issues/1072, I am starting to think that we need to expose MP Config properties for cases when multiple metrics providers are discovered (either by mechanism built-in to SR FT or by the mechanism that WF uses) to handle scenarios when in a single deployment, users elect to use MP Telemetry for tracing, yet want to e.g. continue using Micrometer for MP FT metrics collection. While Telemetry has specific properties for disabling, it would disable for the whole deployment. MP FT has properties for only disabling metrics altogether. While the MP FT spec is alright with collecting metrics with multiple providers simultaneously (which is beneficial for gradual migration or monitoring systems compatibility), there are drawbacks such as performance implications, potentially duplicated metrics if feeding into the same metric. Thoughts?

Ladicek:
Resolves #1078

### DIFF
--- a/doc/modules/ROOT/pages/reference/metrics.adoc
+++ b/doc/modules/ROOT/pages/reference/metrics.adoc
@@ -89,3 +89,19 @@ It is possible to completely disable fault tolerance metrics using MicroProfile 
 ----
 MP_Fault_Tolerance_Metrics_Enabled=false
 ----
+
+It is also possible to disable individual metric providers:
+
+[source,properties]
+----
+# to disable MP Metrics
+smallrye.faulttolerance.mpmetrics.disabled=true
+
+# to disable OpenTelemetry Metrics
+smallrye.faulttolerance.opentelemetry.disabled=true
+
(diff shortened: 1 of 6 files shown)


rhusar:
Let me take this for a quick spin.

Ladicek:
Please report back; I'll keep this open until then.

rhusar (approved):
I tried a few things and this seems to worked exactly as I wanted to. Last piece of the OpenTelemetry puzzle is in place!

Also reviewed the code and LGTM.

rhusar:
As discussed, will be backported to 6.6.x.

Ladicek:
The PR https://github.com/smallrye/smallrye-fault-tolerance/pull/1080 backports this to 6.6. I'll release 6.6.3 later today.

PR #1097, opened 2025-01-03
Improve event loop integration

Ladicek:


Ladicek:
This should fix https://github.com/quarkusio/quarkus/issues/45061

PR #1102, opened 2025-01-08
Update deprecated getMessageLogger(..) method usage

rhusar:


Ladicek (approved):

Ladicek:
I won't merge yet, just in case something comes up and we'll need to do 6.7.3. I have one bigger improvement in the works for 6.8.0, so once that done, I'll merge this too.

rhusar:
Sounds good, not urgent. Just needs to happen sometimes.

PR #1107, opened 2025-02-05
docs: Add io.smallrye:smallrye-fault-tolerance-apiimpl module to the …

rhusar:
…list of required modules since 6.8.0


Ladicek:
Ah, good point! Totally forgot.

Ladicek (approved):

Ladicek:
Thanks!

PR #1126, opened 2025-09-03
Generate `smallrye-fault-tolerance` BOM

gastaldi:
This generates a `io.smallrye:smallrye-fault-tolerance-bom` artifact that can simplify its usage as a managed dependency


gastaldi:
Ahh good catch, forgot to check that. I'll fix it asap

PR #1129, opened 2025-10-06
add Java 25 (and 26-ea)

Ladicek:


Ladicek:
Need to bump Kotlin first.

PR #1275, opened 2026-06-07
fix annotation lookup when FT annotations come from a class-level stereotype

### LINKED ISSUE #1274: @Fallback doesn't see annotations from class-level @Stereotype on the same bean
When using a `@Stereotype` to apply fault tolerance annotations on a class and putting `@Fallback` on a method, retry and circuit breaker stop working entirely.

```java
@MyStereotype  // @Retry(maxRetries=2) + @CircuitBreaker + @Timeout + @Bulkhead
@ApplicationScoped
class MyService {

    public String work() { throw new RuntimeException(); }

    String fallback() { return "fallback"; }
}
```

`work()` is called exactly once regardless of `maxRetries=2`. The circuit breaker never opens either.

Reproduction: https://github.com/TelesNascimento25/smallrye-ft-stereotype-fallback

Tested on Quarkus 3.20.6, 3.34.3 and 3.36.1, found they have same behavior on all three.

TelesNascimento:
Fixes #1274.

When `@Fallback` is at method level and other FT annotations arrive via a class-level `@Stereotype`, `FaultToleranceMethods` returns null for all of them. Neither `AnnotatedType.getAnnotation()` nor `Class.getAnnotation()` look inside `@Stereotype` meta-annotations.

The fix adds a private `findInStereotypes` helper that both the CDI and reflection paths delegate to when a direct lookup returns null. It handles nested stereotypes recursively.

### DIFF
--- a/.github/project.yml
+++ b/.github/project.yml
@@ -1,4 +1,4 @@
 name: SmallRye Fault Tolerance
 release:
-  current-version: 6.11.0
-  next-version: 6.11.1-SNAPSHOT
+  current-version: 6.11.1
+  next-version: 6.11.2-SNAPSHOT
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <groupId>io.smallrye</groupId>
         <artifactId>smallrye-fault-tolerance-parent</artifactId>
-        <version>6.11.1-SNAPSHOT</version>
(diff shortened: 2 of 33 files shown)


TelesNascimento:
Closing to reopen from a clean branch based on main. The fix is correct but the branch was based on the 6.11.1 tag instead of main.

PR #1276, opened 2026-06-07
fix annotation lookup when FT annotations come from a class-level stereotype

### LINKED ISSUE #1274: @Fallback doesn't see annotations from class-level @Stereotype on the same bean

TelesNascimento:
Fixes #1274.

When @Fallback is at method level and other FT annotations arrive via a class-level @Stereotype, FaultToleranceMethods returns null for all of them. Neither AnnotatedType.getAnnotation() nor Class.getAnnotation() look inside @Stereotype meta-annotations.

The fix adds a private findInStereotypes helper that both the CDI and reflection paths delegate to when a direct lookup returns null. It handles nested stereotypes recursively and prevents infinite loops using a visited set.

**CDI vs reflection path note:** The CDI path searches getDeclaringType() directly, while the reflection path walks the superclass hierarchy explicitly. Both are correct: stereotypes marked with @Inherited are visible on subclasses via Java reflection even when a subclass overrides the method, so the CDI path finds inherited stereotypes without walking the hierarchy manually.

**Tests:** 16 tests covering a 4×4 matrix (all using @Retry):
- Axis 1 (annotation placement): stereotype only / class overrides stereotype / method overrides stereotype / class+method override stereotype
- Axis 2 (stereotype discovery): direct on class / inherited from superclass / nested stereotype on class / nested stereotype inherited from superclass

Reproduction: https://github.com/TelesNascimento25/smallrye-ft-stereotype-fallback

### DIFF
--- a/implementation/fault-tolerance/src/main/java/io/smallrye/faulttolerance/config/FaultToleranceMethods.java
+++ b/implementation/fault-tolerance/src/main/java/io/smallrye/faulttolerance/config/FaultToleranceMethods.java
@@ -1,14 +1,20 @@
 package io.smallrye.faulttolerance.config;
 
 import java.lang.annotation.Annotation;
+import java.lang.annotation.Inherited;
 import java.lang.reflect.Method;
 import java.lang.reflect.Type;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.Deque;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
+import jakarta.enterprise.inject.Stereotype;
 import jakarta.enterprise.inject.spi.AnnotatedMethod;
 
 import org.eclipse.microprofile.faulttolerance.Asynchronous;
@@ -80,7 +86,11 @@ public class FaultToleranceMethods {
             directlyPresent.add(annotationType);
             return cdiMethod.getAnnotation(annotationType);
         }
-        return cdiMethod.getDeclaringType().getAnnotation(annotationType);
+        A annotation = cdiMethod.getDeclaringType().getAnnotation(annotationType);
+        if (annotation == null) {
+            annotation = findInStereotypes(cdiMethod.getDeclaringType().getAnnotations(), annotationType);
+        }
+        return annotation;
     }
 
     // ---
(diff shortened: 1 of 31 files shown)


TelesNascimento:
@Retry carries `@Target({METHOD, TYPE})` and `@InterceptorBinding` per the MicroProfile spec. CDI 2.8.1.2 explicitly allows `@Stereotype` to declare interceptor bindings - the spec shows examples with `@Transactional` and `@Secure` on `@Stereotype`-annotated types. Since `@Retry`'s `@Target` includes TYPE, it's valid to place it on a `@Stereotype`.

We've validated this with 16 test scenarios covering composition, inheritance, and conflict resolution. The behavior is solid.

There's a TCK PR ready that formalizes this pattern, maybe worth submitting alongside to establish it as portable across implementations. This closes the gap between what the spec allows and what's actually tested.

Ladicek:
Oh, great point. Forgot about this :-) You're right, this should be possible.

Ladicek:
The fix itself is OK. The tests look impressive, but are actually fairly hard to comprehend. We don't need full "horizontal" coverage (test with all fault tolerance strategies), but full "vertical" coverage (one fault tolerance strategy, but testing with all possible cases). I'm thinking this matrix:

1. axis:
- stereotype that includes FT annotation that is NOT present on a class and is NOT present on a method
- stereotype that includes FT annotation that is present on a class and is NOT present on a method
- stereotype that includes FT annotation that is NOT present on a class and is present on a method
- stereotype that includes FT annotation that is present on a class and is present on a method
2. axis:
- stereotype that is present directly on the class
- stereotype that is inherited from a superclass
- stereotype that is present on another stereotype that is present directly on the class
- stereotype that is present on another stereotype that is inherited from a superclass

This itself leads to a big number of tests (16), so doing it for one FT strategy (say, `@Retry`, that is very easy to test) should be more than enough. I'm fairly confident we don't need to repeat this for all FT strategies.

TelesNascimento:
Reworked the tests as you suggested: 

4×4 matrix with @Retry, 

16 tests covering all combinations of annotation placement (stereotype only / classoverride / method override / class+method) 

and stereotype location (direct / inherited / nested direct / nested inherited).Passing on JDK 17, 21, and 25.

Ladicek:
Thanks! I'll take a look tomorrow.

Ladicek (approved):

Ladicek:
Thanks! I had to modify the logic a little, because while all FT annotations are `@Inherited`, a stereotype doesn't have to be. While at it, I rewrote the logic for searching for transitive stereotypes from recursive to a worklist-based implementation, and slightly renamed the test classes. But again, thanks for spotting the issue!

TelesNascimento:
Thanks for the feedback and the fix! 

The edge case with non-@Inherited stereotypes makes sense, I'll remember that for future
contributions.

PR #1284, opened 2026-06-12
Update to OpenTelemetry 1.62.0 to resolve CVE

jasondlee:
Update artifact version
Add now-needed SecurityManager check

See: https://github.com/advisories/GHSA-rcgg-9c38-7xpx

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,8 +63,8 @@
         <version.microprofile-context-propagation-api>1.3</version.microprofile-context-propagation-api>
         <!-- this is the last Mutiny version that supports Java 11 -->
         <version.mutiny>2.7.0</version.mutiny>
-        <!-- OpenTelemetry 1.48.0 is required by MicroProfile Telemetry 2.1 (MP Platform 7.1) -->
-        <version.opentelemetry>1.48.0</version.opentelemetry>
+        <!-- note that MicroProfile Telemetry 2.1 (MP Platform 7.1) requires 1.48.0  -->
+        <version.opentelemetry>1.62.0</version.opentelemetry>
         <version.opentracing>0.33.0</version.opentracing>
         <version.rxjava3>3.1.12</version.rxjava3>
         <!-- this is the last SmallRye Common version that supports Java 11 -->
@@ -417,6 +417,7 @@
             </dependency>
         </dependencies>
     </dependencyManagement>
+
     <build>
         <plugins>
             <plugin>
@@ -448,7 +449,35 @@
             </plugin>
         </plugins>
     </build>
+
     <profiles>
+        <profile>
+            <id>security-manager</id>
+            <activation>
+                <jdk>(,22]</jdk>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <id>security-manager-test</id>
+                                <phase>test</phase>
+                                <goals>
+                                    <goal>test</goal>
+                                </goals>
+                                <configuration>
+                                    <argLine>-Djava.security.manager -Djava.security.policy=${maven.multiModuleProjectDirectory}/java.policy</argLine>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+
         <profile>
             <id>not-release</id>
             <activation>
--- a/implementation/fault-tolerance/src/main/java/io/smallrye/faulttolerance/metrics/OpenTelemetryProvider.java
+++ b/implementation/fault-tolerance/src/main/java/io/smallrye/faulttolerance/metrics/OpenTelemetryProvider.java
@@ -1,5 +1,7 @@
 package io.smallrye.faulttolerance.metrics;
 
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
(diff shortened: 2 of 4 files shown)


rhusar:
@jasondlee Can you link/share the SM failure? Because if it were the `buildWithCallback` specifically, that we use in multiple places. So just to be sure this will cover all cases.

jasondlee:
> @jasondlee Can you link/share the SM failure? Because if it were the `buildWithCallback` specifically, that we use in multiple places. So just to be sure this will cover all cases.

Sure. Give me a few minutes to reproduce...

jasondlee:
> @jasondlee Can you link/share the SM failure? Because if it were the `buildWithCallback` specifically, that we use in multiple places. So just to be sure this will cover all cases.

Here's the error from the WildFly expansion testsuite:

```
java.security.AccessControlException: WFSM000001: Permission check failed (permission "("java.lang.RuntimePermission" "getClassLoader")" in code source "(vfs:/content/                                    FaultToleranceOpenTelemetryIntegrationTestCase.war/WEB-INF/classes <no signer certificates>)" of "null")
    at org.wildfly.security.elytron-base@2.9.1.Final//org.wildfly.security.manager.WildFlySecurityManager.checkPermission(WildFlySecurityManager.java:341)
    at org.wildfly.security.elytron-base@2.9.1.Final//org.wildfly.security.manager.WildFlySecurityManager.checkPermission(WildFlySecurityManager.java:233)
    at java.base/java.lang.ClassLoader.checkClassLoaderPermission(ClassLoader.java:2060)
    at java.base/java.lang.Thread.getContextClassLoader(Thread.java:1502)
    at io.opentelemetry.sdk@1.62.0//io.opentelemetry.sdk.metrics.internal.state.CallbackRegistration.<init>(CallbackRegistration.java:39)
    at io.opentelemetry.sdk@1.62.0//io.opentelemetry.sdk.metrics.internal.state.CallbackRegistration.create(CallbackRegistration.java:68)
    at io.opentelemetry.sdk@1.62.0//io.opentelemetry.sdk.metrics.InstrumentBuilder.buildLongAsynchronousInstrument(InstrumentBuilder.java:89)
    at io.opentelemetry.sdk@1.62.0//io.opentelemetry.sdk.metrics.SdkLongUpDownCounter$SdkLongUpDownCounterBuilder.buildWithCallback(SdkLongUpDownCounter.java:86)
    at io.smallrye.fault-tolerance@6.11.1//io.smallrye.faulttolerance.metrics.OpenTelemetryProvider.<init>(OpenTelemetryProvider.java:49)
```
This could *probably* use a test in this code base like I did for SmallRye OpenTelemetry. I can work on that.

jasondlee:
@rhusar I've added a SM-enabled surefire execution. The policy file is wide open, but there's no real way to avoid that, given how surefire works and what it needs, but it does exercise the SM branch.

pom.xml
```diff
@@ -63,8 +63,7 @@
         <version.microprofile-context-propagation-api>1.3</version.microprofile-context-propagation-api>
         <!-- this is the last Mutiny version that supports Java 11 -->
         <version.mutiny>2.7.0</version.mutiny>
-        <!-- OpenTelemetry 1.48.0 is required by MicroProfile Telemetry 2.1 (MP Platform 7.1) -->
-        <version.opentelemetry>1.48.0</version.opentelemetry>
+        <version.opentelemetry>1.62.0</version.opentelemetry>
```
Ladicek, line 67: I don't mind the other changes, but what about this one? I know there are breaking changes. What OpenTelemetry version do you use in WildFly? Also, CC @brunobat, what OpenTelemetry version do we use in Quarkus?

brunobat: Quarkus is currently on SDK 1.60.1 and it's also being upgraded to 1.62.0 https://github.com/quarkusio/quarkus/pull/54794

rhusar: @Ladicek We are trying to upgrade OpenTelemetry in WildFly – https://github.com/wildfly/wildfly/pull/20093 – which prompted existence of this PR.

Ladicek: OK, so 1.62.0 looks good. Thanks!

Ladicek (approved):

Ladicek:
I assume you need a release after this gets merged?

jasondlee:
> I assume you need a release after this gets merged?

Yes. That would be very much appreciated. :)

pom.xml
```diff
@@ -448,7 +449,35 @@
...
+                                </goals>
+                                <configuration>
+                                    <argLine>-Djava.security.manager -Djava.security.policy=${maven.multiModuleProjectDirectory}/java.policy</argLine>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
```
rhusar, line 479: Not sure if the recent push fixed this, but what's the idea behind this?

 I tried running the tests WITHOUT the fix and they passed. So this is not an effective test?

pom.xml
```diff
@@ -448,7 +449,35 @@
             </plugin>
         </plugins>
     </build>
+
     <profiles>
+        <profile>
+            <id>security-manager</id>
+            <activation>
+                <jdk>(,22]</jdk>
+            </activation>
```
rhusar, line 458: This activation is problematic in general.

This would mean **always** run with security manager with those JDKs and leaving out the code paths without security managers on those JDKs. This is less of a problem since in CI we test with later JDKs so at least those code paths are then tested on recent JDKs. That's why in general we make the security manager profiles configurable = opt in.

But I ll leave this to maintainer @Ladicek

Ladicek: That's not actually true. This profile _adds_ a Surefire execution with security manager enabled, but the default Surefire execution _doesn't_ have secman enabled, so it is OK.

pom.xml
Ladicek: This is quite possible. If I recall, on the 6.x branch, we actually don't use OpenTelemetry in tests. Well, except perhaps of the TCK, I would expect _that_ to fail I think...

jasondlee: That's a good question, and I really don't know for sure. It _feels_ a little odd as it all it does is exercise the other side of the SecMan check branch, which ultimately runs the same code, just inside the `doPrivileged`. It doesn't seem like it really adds much other than execution time on certain JDKs, but I am 1) trying to be thorough and 2) looking for feedback. Happy to remove this is we think it doesn't really add much.

pom.xml
rhusar: Oh ok, correct about the adding, nevertheless, point stands it's still problematic as it's not controllable, i.e. you may not want to run it for different combinations in e.g. CI or local run. 

Again - I leave that to you.

Ladicek: I think it's fine. I mean, 6.x is effectively dead or will be relatively soon, so I don't really mind :-)

jasondlee: @Ladicek @rhusar Are there any changes you would like me to make? Sounds like the surefire execution is of dubious value, and I'm happy to remove that if you'd like, but is there anything else? Would be nice to have this released today so we can get WF updated...

Ladicek: 6.11.2 should be in Central now.

jasondlee: I see it. Much appreciated.

PR #1319, opened 2026-09-09
Bump OTel to 1.63.0

Ladicek:


Ladicek:
CC @brunobat @jasondlee @rhusar 

This is per @brunobat request for Quarkus 3.x, so I wonder if this is OK for WildFly as well.

rhusar:
@jasondlee How do we usually handle the alignment? I assume this is fine as this will just recompile against later version.

FWIW WildFly is currently on:

> [INFO] +- io.opentelemetry:opentelemetry-api:jar:1.48.0:compile

jasondlee:
> @jasondlee How do we usually handle the alignment? I assume this is fine as this will just recompile against later version.
> 
> FWIW WildFly is currently on:
> 
> > [INFO] +- io.opentelemetry:opentelemetry-api:jar:1.48.0:compile

This is probably fine. It's already well ahead of where WildFly is and seems to run without issue. WildFly will be updating soon as part of the update to MP Telemetry 2.2, so we should see the version gap close, at the very least, if not go away altogether. Worst case scenario, we can branch off a version prior to this update if needed in WildFly.

jasondlee (approved):

Ladicek:
Makes sense, thanks!

PR #1325, opened 2026-09-16
Fix unbounded FT metrics memory leak: key DelegatingMetricsCollector cache by cacheKey()

MIERAI:
## Problem

When a **programmatic** `Guard`/`TypedGuard`/`FaultTolerance` is invoked through `@ApplyGuard`
(or the deprecated `@ApplyFaultTolerance`), Fault Tolerance retains **one
`DelegatingMeteredOperation` + one `MetricsCollector` per invocation, for the lifetime of the
(singleton) guard**. The count grows 1:1 with the number of calls and is never released, so a
long-running, high-throughput service steadily fills the heap (rising GC pressure → slowdown →
OOM). The Micrometer/OpenTelemetry meters themselves stay bounded (de-duplicated by name), which
makes the leak easy to misdiagnose as "needs more memory".

This is a *different* leak from the one fixed in 6.9.0 (that was the UUID method tag when **no**
description is set); this one happens regardless of the description.

## Root cause

`core/.../metrics/DelegatingMetricsCollector`:

```java
private final ConcurrentMap<MeteredOperation, MetricsCollector<V>> cache = new ConcurrentHashMap<>();

public Future<V> apply(FaultToleranceContext<V> ctx) {
    MeteredOperationName name = ctx.get(MeteredOperationName.class);
    MeteredOperation operation = name != null
            ? new DelegatingMeteredOperation(originalOperation, name.get())   // fresh object per call
            : originalOperation;
    ...
    delegate = cache.computeIfAbsent(operation, ...);   // keyed by the INSTANCE
    ...
}
```

On the `@ApplyGuard` / `@ApplyFaultTolerance` path, `FaultToleranceInterceptor` puts a fresh
`MeteredOperationName` into the context on every invocation, so `apply()` builds a new
`DelegatingMeteredOperation` each time. `DelegatingMeteredOperation` has no `equals`/`hashCode`, so
keying the cache by the instance means `computeIfAbsent` never hits — a new `MetricsCollector`
entry is added and retained on every call.

(Declarative annotations like `@Retry` don't put a `MeteredOperationName` in the context, so
`operation == originalOperation` — a stable singleton — and don't leak. The leak is specific to
invoking a programmatic guard via `@ApplyGuard`/`@ApplyFaultTolerance`.)

Note the inconsistency: `MicrometerProvider.create(...)` already caches its `MetricsRecorder` by
`meteredOperation.cacheKey()`, which is why the meters stay bounded while these wrappers do not.

## Fix

Key the cache by `MeteredOperation#cacheKey()` instead of by the instance — making
`DelegatingMetricsCollector` consistent with `MicrometerProvider`:

```diff
-    private final ConcurrentMap<MeteredOperation, MetricsCollector<V>> cache = new ConcurrentHashMap<>();
+    private final ConcurrentMap<Object, MetricsCollector<V>> cache = new ConcurrentHashMap<>();
...
-            delegate = cache.computeIfAbsent(operation,
+            delegate = cache.computeIfAbsent(operation.cacheKey(),
                     ignored -> new MetricsCollector<>(this.delegate, provider.create(operation), operation));
```

## Test

Adds `GeneralMetricsTest.metricsCollectorIsReusedForSameOperation`: it drives
`DelegatingMetricsCollector.apply()` 1000 times, each with a fresh `MeteredOperationName` (as the
interceptor does), and asserts the `MetricsProvider` is asked to create a collector only **once**.

- Before this change: `createCount == 1000` (a collector per invocation) → test fails.
- After this change: `createCount == 1` → test passes.

## Affected versions

The affected classes are byte-for-byte identical in 6.9.x, 6.10.x, 6.11.x and 7.0.0-RC1, so all
are affected. Verified against 6.11.3.

## Notes

- An alternative fix is to give `DelegatingMeteredOperation` an `equals`/`hashCode` delegating to
  `cacheKey()`; keying the map by `cacheKey()` is the smaller, more local change and keeps both
  caches consistent. Happy to adjust to whichever you prefer.
- A standalone end-to-end reproducer (plain CDI / Weld SE, `@ApplyGuard`, no description) is
  available if useful; the leak there shows as `DelegatingMeteredOperation` retained == number of
  invocations.


Ladicek:
Amazing finding, thanks a lot! The fix is correct. I'll apply some small tuning (mostly add an actual reproducing test) and merge.

Ladicek (approved):

yousaikei:
@Ladicek 
Thanks so much for the quick turnaround and for adding the reproducing test! For context, we hit this in production across ~40 services (Quarkus 3.27 / SmallRye FT 6.9.x): a long-lived, high-traffic queue processor slowly filled the heap toward OOM. Since the fix landed in the 7.0.0 milestone — would you consider backporting it to a 6.x release (e.g. 6.11.x/6.12) as well? We use the programmatic Guard + @ApplyGuard pattern on 6.x and can't move to 7.0 easily yet. Thanks again!

Ladicek:
I already have a backport PR to 6.11 (see #1327) and we're handling additional backports internally.

MIERAI:
That's perfect — 6.11.x is exactly the line we're on, so we'll pick it up as soon as it's released. Really appreciate the fast turnaround. Thanks again!

Ladicek:
You mentioned above you're on 6.9, but I also did backports to 6.10 (#1328) and 6.9 (#1329) and will release them shortly.

MIERAI:
Wow — a 6.9 backport is more than I hoped for. That lets us pick up the real fix with zero version jump. Really generous of you, thank you so much! 🙏 We'll upgrade to the patched 6.9.x as soon as it's out.

PR #1329, opened 2026-09-16
[6.9] fix memory leak in metrics: key DelegatingMetricsCollector cache by cacheKey()

Ladicek:
`DelegatingMetricsCollector` used to cache the `MetricsCollector` instances in a `Map` keyed by the `MeteredOperation` instance. This is wrong, because no implementation of `MeteredOperation` defines `equals()` / `hashCode()`.

This only matters in one situation: using the programmatic API (`Guard` or `TypedGuard`) in a declarative manner (`@ApplyGuard`). This is because:

- only the programmatic API uses `DelegatingMetricsCollector`
- only the declarative manner uses custom `MeteredOperationName`

In this case, each execution would put a new entry to the cache, effectively leaking memory. Fortunately, `MeteredOperation` already has a `cacheKey()` method we can use here (it's used elsewhere too). This is what this commit does, avoiding the memory leak.


MIERAI:
thank you so much!

